PR #1, opened 2013-08-20
SONARPLUGINS-3116

kalidasya:
Add support for django (dtl) templates. (https://docs.djangoproject.com/en/dev/topics/templates/)
I also fixed a minor issue, the init() was never called in the HtmlAstScanner.


fmallet:
Relating issue ticket: https://jira.sonarsource.com/browse/SONARWEB-17

m-g-sonar:
this implementation is outdated. Also, due to the lack of traction, we won't merge it. However, we keep the ticket in our backlog.

PR #3, opened 2015-05-23
Fixes for SONARWEB-25

renewolfert:
Fixed issue with processing of value of unquoted attributes by replacing the use of the EndTokenMatcher by reading from the CodeReader until the end of the value is detected


mpaladin:
@renewolfert thank you for your contribution, the fix was not passing ITs and was failing some corner cases, I merged your contribution with few adjustments in https://github.com/SonarSource/sonar-web/pull/20: https://github.com/SonarSource/sonar-web/commit/c46b20098d7f859d270e1dd82f39d082dd249808.

PR #14, opened 2016-05-29
new rule template for detecting a regular expression pattern on a single line

boyarsky:
I needed a rule that checks web files for arbitrary strings on a line. They could be anywhere in the file and a substring of attributes, character data, etc.  I described the use case in [my blog](https://www.selikoff.net/2016/05/29/sonar-plugins-and-why-i-had-to-fork-the-sonar-web-plugin-to-add-a-rule/). Since the rule has to be in the web plugin (based on file type and not being able to extend the web plugin externally), I forked to add the rule template. I made it a rule template to make it more generic.

Submitting a pull request in case you are interested in incorporating this rule template in the actual sonar web plugin.


fmallet:
Following ticket was created to track this request for feature: https://jira.sonarsource.com/browse/SONARWEB-76.

saberduck:
hello @boyarsky,

while I understand that this rule served your need, I do not believe that this is generally useful, because regex doesn't respect the structure of html. Also we are trying to limit the usage of template rules, because of some of their limitations.
I am thus closing the PR and the corresponding ticket.

PR #39, opened 2018-02-16
SONARWEB-54 rename plugin

m-g-sonar:


sonar-web-plugin/pom.xml
```diff
@@ -7,20 +7,21 @@
...
   </parent>
-  
+
   <artifactId>sonar-web-plugin</artifactId>
   <packaging>sonar-plugin</packaging>
 
-  <name>SonarSource :: Web :: Sonar Plugin</name>
-  <description>Enables scanning of HTML, and JSP/JSF files.</description>
+  <name>SonarWeb :: SonarQube Plugin</name>
+  <description>Code analyzer for Web (HTML, JSP, JSF, ..)</description>
```
ganncamp: No need for ellipses. There are really only 3 in the list, currently.

PR #50, opened 2018-05-04
Add rule to check nested script nodes.

dodie:
This rule attempts to find nested script tags in the HTML code. Nested script tags are problematic,
because the browsers does not support them, treating the contents of the tag as plain text, and closing it at the first occurrencce of the `</script>` charcter sequence.

Please see these SO entries for more examples:

- https://stackoverflow.com/questions/6322541/nested-javascript-templates-is-this-possible-valid
- https://stackoverflow.com/questions/31773709/nested-script-tags-comparison-of-2-approaches
- https://stackoverflow.com/questions/12500613/in-html5-spec-it-seems-the-nested-script-tags-are-supported-so-what-does-a-rea

`ChildElementIllegalCheck` can't be parameterized to detect nested script nodes,
as the contents of a script node is interpreted as text (so there are no child nodes from Sonar Web model point of view).
And even if it would notice the child nodes in scripts, the ChildElementIllegalCheck only check immediete child nodes,
so it would not create violations for `script > div > script` hierarchies.

### DIFF
--- a/its/plugin/src/test/resources/com/sonar/it/web/backup.xml
+++ b/its/plugin/src/test/resources/com/sonar/it/web/backup.xml
@@ -79,6 +79,11 @@
     </rule>
     <rule>
       <repositoryKey>Web</repositoryKey>
+      <key>NestedJavaScriptCheck</key>
+      <priority>CRITICAL</priority>
+    </rule>
+    <rule>
+      <repositoryKey>Web</repositoryKey>
       <key>MultiplePageDirectivesCheck</key>
       <priority>MINOR</priority>
     </rule>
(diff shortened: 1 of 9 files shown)


fmallet:
Thanks a lot @dodie for this contribution. I've created the following relating rule specification and implementation ticket:
* https://jira.sonarsource.com/browse/RSPEC-4645
* https://jira.sonarsource.com/browse/SONARWEB-75

The PR will be merged part of our next sprint on SonarWeb.

benzonico:
@dodie  Thanks for the contribution : I opened to trigger our internal QA and update a bit the implementation and rule metadatas : https://github.com/SonarSource/sonar-html/pull/58

As a consequence I am closing this PR. Thanks again.

PR #51, opened 2018-05-07
Drop dependency on slf4j

dbmeneses:


benzonico:
fixed via https://github.com/SonarSource/sonar-html/pull/65 Thanks @dbmeneses for this patch.

PR #58, opened 2018-07-31
SONARWEB-75 implements RSPEC-4645 about nested script tags

benzonico:


sonar-web-plugin/src/main/resources/org/sonar/l10n/web/rules/Web/NestedJavaScriptCheck.html
```diff
@@ -0,0 +1,33 @@
+<p>When parsing a script node, the browser treats its contents as plain text, and immediately finishes parsing when it finds the first closing <code>&lt;/script&gt;</code> character sequence.
```
pynicolas: Why this file? We already have S4645.html

PR #70, opened 2018-08-02
 SONARHTML-87 Remove dependency on sslr-squid-bridge

pynicolas:


sonar-html-plugin/pom.xml
```diff
@@ -89,16 +74,16 @@
...
+      <groupId>org.sonarsource.analyzer-commons</groupId>
+      <artifactId>sonar-analyzer-recognizers</artifactId>
+      <version>${analyzerCommons.version}</version>
+    </dependency>
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
       <version>15.0</version>
     </dependency>
-    <dependency>
```
benzonico: This is used by IllegalElementCheck (among many others) and should not be removed.

pynicolas: Fixed in 4581e29

PR #72, opened 2018-08-02
SONARHTML-43 Fix NPE in DeprecatedAttributesInHtml5Check

pynicolas:


benzonico:
probably worth adding the testcase in `IllegalAttributeCheck` otherwise LGTM

PR #74, opened 2018-08-02
SONARHTML-1 update counting of ncloc

benzonico:


sonar-html-plugin/src/main/java/org/sonar/plugins/html/analyzers/PageCountLines.java
```diff
@@ -43,13 +43,11 @@
...
   private int headerCommentLines;
-  private int linesOfCode;
   private final Set<Integer> detailedLinesOfCode = Sets.newHashSet();
   private final Set<Integer> detailedLinesOfComments = Sets.newHashSet();
 
   @Override
   public void startDocument(List<Node> nodes) {
-    linesOfCode = 0;
     blankLines = 0;
     commentLines = 0;
```
pynicolas: I think we should do the same thing for comment lines. One more thing: blank lines inside comments should probably not be counted as comment lines.

PR #78, opened 2018-08-03
SONARHTML-16 ignore html starting tags and doctype from duplication

benzonico:


sonar-html-plugin/src/test/java/org/sonar/plugins/html/core/HtmlSensorTest.java
```diff
@@ -107,6 +107,21 @@ public void testSensor() throws Exception {
...
+  @Test
+  public void should_not_count_doctype_and_html_tag_in_duplication() throws Exception {
+    DefaultInputFile inputFile = createInputFile(TEST_DIR, "create-salesorder_duplication.xhtml");
+    tester.fileSystem().add(inputFile);
+
+    sensor.execute(tester);
+
+    String componentKey = inputFile.key();
+    // 206 tokens with no filtering
+    assertThat(tester.cpdTokens(componentKey)).hasSize(196);
```
pynicolas: Given the limited computing capabilities of my brain, I would have preferred a smaller test case.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/core/HtmlTokensVisitor.java
```diff
@@ -71,7 +72,16 @@ private void highlightAndDuplicate() {
       TokenType tokenType = token.getType();
       if (!tokenType.equals(GenericTokenType.EOF)) {
         TokenLocation tokenLocation = new TokenLocation(token);
-        cpdTokens.addToken(tokenLocation.startLine(), tokenLocation.startCharacter(), tokenLocation.endLine(), tokenLocation.endCharacter(), token.getValue());
+        String tokenValue = token.getValue();
+        if(tokenValue.startsWith("<html")) {
+          htmlTag = true;
+        }
+        if(!tokenValue.startsWith("<!DOCTYPE") && !htmlTag) {
+          cpdTokens.addToken(tokenLocation.startLine(), tokenLocation.startCharacter(), tokenLocation.endLine(), tokenLocation.endCharacter(), tokenValue);
```
pynicolas: Instead of filtering out some tokens, it would be better ignore all tokens before the end of the "header". If we don't do that, we may create some very strange duplications. For the following files, your code would create a duplication starting from line 1, even if lines 2 are different:
file1:
```
<?xml version='1.0' encoding='UTF-8' ?> 
<html xmlns="http://www.w3.org/1999/xhtml">
<Big duplicated block>
```

file2:
```
<?xml version='1.0' encoding='UTF-8' ?> 
<html xmlns="completely different but ignored">
<Big duplicated block>
```

PR #83, opened 2018-08-06
SONARHTML-74 analyze php extensions files

benzonico:


sonar-html-plugin/src/main/java/org/sonar/plugins/html/core/HtmlSensor.java
```diff
@@ -88,8 +90,11 @@ public void execute(SensorContext sensorContext) {
     Iterable<InputFile> inputFiles = fileSystem.inputFiles(
       predicates.and(
         predicates.hasType(InputFile.Type.MAIN),
-        predicates.hasLanguage(HtmlConstants.LANGUAGE_KEY))
-    );
+        predicates.or(
+          predicates.hasLanguage(HtmlConstants.LANGUAGE_KEY),
+          predicates.or(Stream.of(PHP_FILE_SUFFIXES).map(predicates::hasExtension).toArray(FilePredicate[]::new))
```
alban-auzeill: why maintaining a hard-coded list of php suffixes (PHP_FILE_SUFFIXES) and not just use the following predicate ?
`predicates.hasLanguage("php")`

benzonico: This won't work if php plugin is not installed.

PR #84, opened 2018-08-06
SONARHTML-63 Rule: Track uses of "FIXME" tags

alban-auzeill:


sonar-html-plugin/src/test/java/org/sonar/plugins/html/checks/comments/CommentUtilsTest.java
```diff
@@ -0,0 +1,38 @@
...
+import static org.sonar.plugins.html.checks.comments.CommentUtils.lineNumber;
+
+public class CommentUtilsTest {
+
+  @Test
+  public void name() {
+    CommentNode node = new CommentNode();
+    node.setStartLinePosition(1);
+    node.setCode("<!--A\nB\nC-->");
+    assertThat(lineNumber(node, 4)).isEqualTo(1);
```
benzonico: What if I pass -1 as offset to CommentNode ?

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/comments/CommentUtils.java
```diff
@@ -0,0 +1,39 @@
...
+
+final class CommentUtils {
+
+  private CommentUtils() {
+    // utlility
+  }
+
+  static int lineNumber(CommentNode node, int offset) {
+    int lineNumber = node.getStartLinePosition();
+    String contentBeforeTodo = node.getCode().substring(0, offset);
```
benzonico: Name of the variable is incorrect in this case now.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/comments/CommentUtils.java
```diff
@@ -0,0 +1,39 @@
...
+    for (int i = 0; i < contentBeforeTodo.length(); i++) {
```
benzonico: A way to write this could be `content.chars().filter(c ->  c == '\n').count()`

alban-auzeill: I agree, it looks more concise, I have rewritten the method.

sonar-html-plugin/src/test/java/org/sonar/plugins/html/checks/comments/CommentUtilsTest.java
benzonico: Please add a proper invalidation of argument.

alban-auzeill: done

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/comments/CommentUtils.java
```diff
@@ -26,6 +26,9 @@ private CommentUtils() {
   }
 
   static int lineNumber(CommentNode node, int offset) {
+    if (offset < 0) {
+      throw new IllegalArgumentException("Negative offset: " + offset);
```
benzonico: Message is meaningless and untested. Please have a message that will help another developer looking at the stacktrace understand the issue.

PR #85, opened 2018-08-06
SONARHTML-77 Rule: Video should have subtitles

alban-auzeill:


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/VideoTrackCheck.java
```diff
@@ -0,0 +1,60 @@
...
+    return node.getAttributes().stream().anyMatch(VideoTrackCheck::isSrcAttribute) ||
+      node.getChildren().stream().anyMatch(VideoTrackCheck::isSourceTag);
+  }
+
+  private static boolean hasTrackChild(TagNode node) {
+    return node.getChildren().stream().anyMatch(VideoTrackCheck::isTrackTag);
+  }
+
+  private static boolean isVideoTag(TagNode node) {
+    return "VIDEO".equalsIgnoreCase(node.getNodeName());
```
benzonico: please use node.equalsElementName which will cover the usage of local name as well.

PR #91, opened 2018-08-10
InlineStyleCheck: The compliant solution was non-compliant

thepieterdc:
The compliant solution still used the `style` attribute.


thepieterdc:
Is fixed in #4181edb5

PR #100, opened 2019-01-14
SONARHTML-94 Declare JSP langage

benzonico:


sonar-html-plugin/src/main/java/org/sonar/plugins/html/HtmlPlugin.java
```diff
@@ -60,13 +64,21 @@ public void define(Context context) {
...
         .description("List of file suffixes that will be scanned.")
         .category(CATEGORY)
         .defaultValue(HtmlConstants.FILE_EXTENSIONS_DEF_VALUE)
         .deprecatedKey("sonar.web.file.suffixes")
         .onQualifiers(Qualifiers.PROJECT)
         .multiValues(true)
+        .build(),
+      PropertyDefinition.builder(HtmlConstants.JSP_FILE_EXTENSIONS_PROP_KEY)
+        .name("JSP File suffixes")
+        .description("List of file suffixes that will be scanned.")
```
pynicolas: This description is identical to the the "HTML File suffixes" property: I think it would be better to make them different.

PR #102, opened 2019-01-16
SONARHTML-92 FP and FN from multiple rules when using Angular's property bindings

christophe-zurn-sonarsource:


sonar-html-plugin/src/main/java/org/sonar/plugins/html/node/TagNode.java
```diff
@@ -49,8 +47,35 @@ public boolean equalsElementName(String elementName) {
...
   }
 
-  public String getAttribute(String attributeName) {
+  /**
+   *  This method takes into account the property binding mechanism of angular. See SONARHTML-92
+   */
+  @Nullable
+  public Attribute getProperty(String propertyName) {
+    String angularProperty = "[" + propertyName + "]";
+    for (Attribute a : attributes) {
```
benzonico: matter of style (so up to you to modify it or keep it as is) : 
`attributes.stream().map(Attribute::getName).filter(...).findFirst().orElse(null);`

christophe-zurn-sonarsource: Ok. I'll keep it like this here then, to have consistent style inside the same file

christophe-zurn-sonarsource:
Indeed. I'll be more careful for next ticket involving many changes

PR #105, opened 2019-03-01
Correcting inline style attribute check

r4d:
The check is only failing on nodes named "<style>" in markup instead of styles inline as attribute as the rule specifically says it is checking.  This update repairs the check to look for an attribute and not a node name.

As the explanation says:
Noncompliant Code Example
< body >
  < h1 style="color: blue;" >Hello World!< /h1 >


Alisamix:
This should definitely be merged asap.

r4d:
lol, open source my a**

PR #115, opened 2019-08-02
SONARHTML-100 Rule S5257: HTML "<table>" should not be used for layout purposes

yassin-kammoun-sonarsource:


sonar-html-plugin/src/test/resources/checks/LayoutTableCheck.html
```diff
@@ -0,0 +1,29 @@
...
+  <tr>
+    <td></td>
+  </tr>
+</table>
+
+<table role="none"> <!-- Non-Compliant -->
+  <tr>
+    <td></td>
+  </tr>
+</table>
```
christophe-zurn-sonarsource: You could add a case with uppercase or mixed letters `role="NONE"`, as well as another with custom tag library prefix to show it does not raise an issue.

PR #117, opened 2019-08-02
SONARHTML-101 Rule S5258: Tables used for layout should not include semantic markup

yassin-kammoun-sonarsource:


sonar-html-plugin/src/main/resources/org/sonar/l10n/web/rules/Web/S5258.html
```diff
@@ -0,0 +1,36 @@
+<p>Tables used for layout should not include semantic markup, such as <code>&lt;th&gt;</code> elements, as it can confuses assistive technologies. At
+best this information will be ignored by screen readers and makes the code less maintainable. However it can also confuse some screen readers and
+reduce the web page accessibility.</p>
+<p>This rule raises an issue when a <code>&lt;table&gt;</code> element containing the&nbsp;<code>role</code> attribute set to
+<code>"presentation"</code> or <code>"none"</code> also contains any of:</p>
+<p> * a <code>&lt;caption&gt;</code> element</p>
+<p> * a <code>&lt;th&gt;</code> element</p>
+<p> * a non-empty <code>summary</code> attribute</p>
+<p> * an <code>&lt;td&gt;</code> element with a <code>headers</code> or <code>scope</code> attribute</p>
```
christophe-zurn-sonarsource: Can you please regenerate rules metadata. I updated jira ticket description so that markup will be correctly generated. Thanks

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/LayoutTableWithSemanticMarkupCheck.java
```diff
@@ -0,0 +1,82 @@
...
+
+import org.sonar.check.Rule;
+import org.sonar.plugins.html.checks.AbstractPageCheck;
+import org.sonar.plugins.html.node.Attribute;
+import org.sonar.plugins.html.node.TagNode;
+
+@Rule(key = "S5258")
+public class LayoutTableWithSemanticMarkupCheck extends AbstractPageCheck {
+
+  private boolean isWithinLayoutTable = false;
```
christophe-zurn-sonarsource: I think you should use a `Deque<Boolean>` to keep whether last `table` element seen was a layout table in this case, otherwise we might have issue with nested tables.
Current implementation will have the following FP/FN:
```
<table role="presentation">
  <caption></caption> <!-- Noncompliant -->
  <tr>
    <td>first</td>
    <td>
      <table>
        <caption></caption> <!-- False positive, this caption is not related to a layout table -->
        <tr><td>nested table</td></tr>
        <tr><td>nested table</td></tr>
      </table>
    </td>
  </tr>
  <tr>
    <td></td>
    <td headers="h"></td> <!-- False negative -->
    <td scope="s"></td> <!-- False negative -->
  </tr>
</table>
```

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/LayoutTableWithSemanticMarkupCheck.java
```diff
@@ -0,0 +1,82 @@
...
+      findAttribute(node, "SUMMARY").ifPresent(a -> createViolation(a.getLine(), format("Remove this \"%s\" attribute", a.getName())));
+      isWithinLayoutTable = true;
+    }
+    if (isWithinLayoutTable) {
+      if (isCaption(node) || isTableHeader(node)) {
+        createViolation(node.getStartLinePosition(), format("Remove this \"%s\" element", node.getNodeName()));
+      }
+      if (isTableColumn(node)) {
+        findAttribute(node, "HEADERS").ifPresent(a -> createViolation(a.getLine(), format("Remove this \"%s\" attribute", a.getName())));
+        findAttribute(node, "SCOPE").ifPresent(a -> createViolation(a.getLine(), format("Remove this \"%s\" attribute", a.getName())));
```
christophe-zurn-sonarsource: Minor: this is more a matter of personal preference, but IMO I'd rather have lambdas slighly longer but with meaningful parameter names, such as `attribute` instead of `a`. But as this is a matter of preference, you can leave it as is if you prefer

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/LayoutTableWithSemanticMarkupCheck.java
```diff
@@ -0,0 +1,96 @@
...
+            createViolation(attribute.getLine(), format("Remove this \"%s\" attribute", attribute.getName())));
+      } else {
+        isWithinLayoutTable.addFirst(false);
+      }
+    }
+    if (!isWithinLayoutTable.isEmpty() && Boolean.TRUE.equals(isWithinLayoutTable.peekFirst())) {
+      if (isCaption(node) || isTableHeader(node)) {
+        createViolation(node.getStartLinePosition(), format("Remove this \"%s\" element", node.getNodeName()));
+      }
+      if (isTableColumn(node)) {
```
christophe-zurn-sonarsource: You could use an `else if` here as the conditions are mutually exclusives

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/LayoutTableWithSemanticMarkupCheck.java
```diff
@@ -0,0 +1,96 @@
...
+        findAttribute(node, "HEADERS").ifPresent(attribute ->
+            createViolation(attribute.getLine(), format("Remove this \"%s\" attribute", attribute.getName())));
+        findAttribute(node, "SCOPE").ifPresent(attribute ->
```
christophe-zurn-sonarsource: I just realised that `findAttribute` is used the same way in all 3 places. You could update update the method to raise issue, for instance `raiseViolationOnAttribute(node, "attributeName")`

PR #120, opened 2019-08-05
SONARHTML-99 Rule S5256: Tables should have headers

yassin-kammoun-sonarsource:


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableWithoutHeaderCheck.java
```diff
@@ -0,0 +1,91 @@
...
+    boolean hasHeader;
+
+    public TableEntry(int line, boolean ignored) {
+      this.line = line;
+      this.ignored = ignored;
+      this.hasHeader = false;
+    }
+  }
+
+  private Deque<TableEntry> tables = new LinkedList<>();
```
christophe-zurn-sonarsource: I feel this solution using `Deque` and keeping states of seen `table` elements is a bit overly complicated.
You could get the same result by checking `if (isTable(node) && node.getChildren().stream().noneMatch(TableWithoutHeaderCheck::isTableHeader))` inside the `startElement` method, and checking recusively for `aria-hidden` attribute of parent nodes

sonar-html-plugin/src/test/resources/checks/TableWithoutHeaderCheck.html
```diff
@@ -0,0 +1,63 @@
...
+<table aria-hidden="false"> <!-- Non-Compliant -->
+  
+</table>
+
+<table aria-hidden="true">
+  <table></table>
+</table>
+
+<table aria-hidden="true">
+  <table aria-hidden="false"></table>
```
christophe-zurn-sonarsource: I'm not sure about this. But wouldn't setting `aria-hidden` attribute explicitely to `false` here override the parent aria-hidden attribute ?

yassin-kammoun-sonarsource: According to the [Accessibility Developer Guide](https://www.accessibility-developer-guide.com/examples/sensible-aria-usage/hidden/), it will not be overridden.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableWithoutHeaderCheck.java
yassin-kammoun-sonarsource: I relied on `Deque` because I wrongly considered `<table>` having headers as long as it has a `<th>` somewhere. However, it seems that headers can only be located within a `<tr>`. I will change that then.

Regarding the `aria-hidden` thing, I like that idea; it may be a bit expensive but definitely more elegant.

PR #121, opened 2019-08-06
SONARHTML-105 Rule S5264: "<object>" tags should provide an alternati…

yassin-kammoun-sonarsource:
…ve content


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/ObjectWithAlternativeContentCheck.java
```diff
@@ -0,0 +1,86 @@
...
+
+  @Override
+  public void endDocument() {
+    stack.clear();
+  }
+
+  @Override
+  public void startElement(TagNode node) {
+    if (!stack.isEmpty()) {
+      stack.peekFirst().hasAlternativeContent = true;
```
saberduck: why not `peek()`?

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/ObjectWithAlternativeContentCheck.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+    if (isObject(node)) {
+      stack.addFirst(new ObjectEntry(node));
```
saberduck: I prefer to use "stack" verbs if we work with stack and don't use feature of deque. There are  `push`/`pop` methods available

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/ObjectWithAlternativeContentCheck.java
```diff
@@ -0,0 +1,86 @@
...
+    TagNode object;
+    boolean hasAlternativeContent;
+    
+    public ObjectEntry(TagNode object) {
+      this.object = object;
+      this.hasAlternativeContent = false;
+    }
+  }
+
+  private Deque<ObjectEntry> stack = new LinkedList<>();
```
saberduck: I am not sure you actually need a stack of `object` elements to be able to report issue. It seems that you only need stack which is 1 element deep, so you can just keep a field, where you set `null` in `endElement`. It will be bit simpler that way.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/ObjectWithAlternativeContentCheck.java
yassin-kammoun-sonarsource: We can indeed use `peek()` instead. For some reason, I got confused by `Deque`'s documentation.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/ObjectWithAlternativeContentCheck.java
yassin-kammoun-sonarsource: Same reason as above.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/ObjectWithAlternativeContentCheck.java
yassin-kammoun-sonarsource: Indeed. The stack was actually useful for an older version of the spec.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/ObjectWithAlternativeContentCheck.java
```diff
@@ -0,0 +1,84 @@
...
+import org.sonar.check.Rule;
+import org.sonar.plugins.html.checks.AbstractPageCheck;
+import org.sonar.plugins.html.node.Node;
+import org.sonar.plugins.html.node.TagNode;
+import org.sonar.plugins.html.node.TextNode;
+
+@Rule(key = "S5264")
+public class ObjectWithAlternativeContentCheck extends AbstractPageCheck {
+
+  private static class ObjectTag {
```
saberduck: You don't need this class, it's enough to have a field to keep the state

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/ObjectWithAlternativeContentCheck.java
```diff
@@ -0,0 +1,84 @@
...
+  private ObjectTag object;
+
+  @Override
+  public void startDocument(List<Node> nodes) {
+    object = null;
+  }
+
+  @Override
+  public void endDocument() {
+    object = null;
```
saberduck: Is this override necessary?

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/ObjectWithAlternativeContentCheck.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+    if (isObject(node)) {
+      object = new ObjectTag(node);
+    }
+  }
+
+  @Override
+  public void endElement(TagNode node) {
+    if (isObject(node) && object != null) {
+      if (!object.hasAlternativeContent) {
```
saberduck: Can `object` be not null and `hasAlternativeContent == false` ? You could probably set `object` to `null` whenever you set `hasAlternativeContent = true`, that way you won't need the class to keep the state.

PR #123, opened 2019-08-06
SONARHTML-98 Rule S5255: "aria-label" or "aria-labelledby" attributes…

yassin-kammoun-sonarsource:
… should be used to differentiate similar elements

### DIFF
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/IndistinguishableSimilarElementsCheck.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/IndistinguishableSimilarElementsCheck.java
@@ -81,7 +81,7 @@ public class IndistinguishableSimilarElementsCheck extends AbstractPageCheck {
     Map<String, List<TagNode>> matched = nodes.stream().collect(Collectors.groupingBy(node -> node.getAttribute("ROLE").toUpperCase(Locale.ROOT)));
     for (List<TagNode> matches : matched.values()) {
       List<TagNode> labeless = matches.stream().filter(match -> !hasAriaLabel(match)).collect(Collectors.toList());
-      if (labeless.size() > 1) {
+      if (labeless.size() > 1 || (!labeless.isEmpty() && labeless.size() != matches.size())) {
         for (TagNode node : labeless) {
           createViolation(node.getStartLinePosition(), "Add an \"aria-label\" or \"aria-labbelledby\" attribute to this element.");
         }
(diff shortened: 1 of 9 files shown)


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/IndistinguishableSimilarElementsCheck.java
```diff
@@ -0,0 +1,107 @@
...
+    } else if (isAside(node)) {
+      asides.add(node);
+    } else if (hasLandmarkRole(node)) {
+      elements.add(node);
+    }
+  }
+
+  private void raiseViolationOnMissingAriaLabel(List<TagNode> nodes) {
+    List<TagNode> matches = nodes.stream().filter(node -> !hasAriaLabel(node)).collect(Collectors.toList());
+    raiseViolationOnSize(matches);
```
christophe-zurn-sonarsource: The way I understand the RSPEC is that as long as there are more than 1 `nav`/`aside` tag, each one should have an aria label. The way it currently works, if there are several `nav` tags but only 1 without the label, it will not raise the issue

sonar-html-plugin/src/test/resources/checks/IndistinguishableSimilarElementsCheck/MultipleNavAside.html
```diff
@@ -0,0 +1,60 @@
...
+<div role="main"> <!-- Compliant -->
+</div>
+
+<div role="navigation"> <!-- Compliant -->
+</div>
+
+<div role="search"> <!-- Compliant -->
+</div>
+
+<div role="application"> <!-- Compliant -->
```
christophe-zurn-sonarsource: I would have expected this one to be non-compliant since there is another element with role "application" and this div element does not have an aria label property

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/IndistinguishableSimilarElementsCheck.java
```diff
@@ -81,7 +81,7 @@ private void raiseViolationOnDuplicateLandmarkRole(List<TagNode> nodes) {
     Map<String, List<TagNode>> matched = nodes.stream().collect(Collectors.groupingBy(node -> node.getAttribute("ROLE").toUpperCase(Locale.ROOT)));
     for (List<TagNode> matches : matched.values()) {
       List<TagNode> labeless = matches.stream().filter(match -> !hasAriaLabel(match)).collect(Collectors.toList());
-      if (labeless.size() > 1) {
+      if (labeless.size() > 1 || (!labeless.isEmpty() && labeless.size() != matches.size())) {
```
christophe-zurn-sonarsource: Note that `matches.size() > 1` condition would yield the same result. I let you decide whether you prefer to keep as is or not.

yassin-kammoun-sonarsource: Well spotted! I will change that.

PR #126, opened 2019-08-07
SONARHTML-99 Fix search for table headers

yassin-kammoun-sonarsource:


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableWithoutHeaderCheck.java
```diff
@@ -46,12 +46,11 @@ private static boolean isHidden(TagNode node) {
...
-        row -> row.getChildren().stream().anyMatch(TableWithoutHeaderCheck::isTableHeader));
-  }
-
-  private static boolean isTableRow(TagNode node) {
-    return "TR".equalsIgnoreCase(node.getNodeName());
+    if (node.getChildren().stream().anyMatch(TableWithoutHeaderCheck::isTableHeader)) {
+      return true;
+    } else {
+      return node.getChildren().stream().filter(child -> !isTable(child)).anyMatch(TableWithoutHeaderCheck::hasHeader);
+    }
```
christophe-zurn-sonarsource: You could use shortcircuit evaluation to avoid `if-else` block here:
```
return node.getChildren().stream().anyMatch(TableWithoutHeaderCheck::isTableHeader) ||
    node.getChildren().stream().filter(child -> !isTable(child)).anyMatch(TableWithoutHeaderCheck::hasHeader);
```

PR #129, opened 2019-08-09
SONARHTML-110 Accept aria-label, aria-labelledby and implicit labels

saberduck:


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/InputWithoutLabelCheck.java
```diff
@@ -39,29 +41,40 @@
...
         return;
       }
-      String id = node.getAttribute("id");
+      String id = getNodeId(node);
 
       if (id == null) {
         createViolation(node.getStartLinePosition(), "Add an \"id\" attribute to this input field and associate it with a label.");
       } else {
-        inputIdToLine.put(node.getAttribute("id"), node.getStartLinePosition());
+        inputIdToLine.put(getNodeId(node), node.getStartLinePosition());
```
yassin-kammoun-sonarsource: You can use `id` here instead of calling again `getNodeId(node)`.

PR #131, opened 2019-08-12
SONARHTML-102 Rule S5260: Table cells should reference their headers

yassin-kammoun-sonarsource:


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderReferenceCheck.java
```diff
@@ -0,0 +1,398 @@
...
+  private void raiseViolationOnInvalidReference(Table table) {
+    Map<TagNode, List<String>> referenceableHeaders = findReferenceableHeadersPerColumnNode(table);
+    for (int row = 0; row < table.rows().size(); ++row) {
+      for (int col = 0; col < table.rows().get(row).size(); ++col) {
+        Table.Column column = table.rows().get(row).get(col);
+        if (column != null) {
+          TagNode node = column.node();
+          if (hasHeaders(node)) {
+            List<String> actual = getHeaders(column.node());
+            List<String> expected = referenceableHeaders.getOrDefault(node, new ArrayList<>());
```
saberduck: you can use `Collections.emptyList` to provide default value

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderReferenceCheck.java
```diff
@@ -0,0 +1,398 @@
...
+  @Override
+  public void endElement(TagNode node) {
+    if (isTable(node) && !stack.isEmpty()) {
+      raiseViolationOnInvalidReference(stack.pop().build());
+    }
+  }
+
+  private void raiseViolationOnInvalidReference(Table table) {
+    Map<TagNode, List<String>> referenceableHeaders = findReferenceableHeadersPerColumnNode(table);
+    for (int row = 0; row < table.rows().size(); ++row) {
```
saberduck: This loop pattern
```
 for (int row = 0; row < table.rows().size(); ++row) {
      for (int col = 0; col < table.rows().get(row).size(); ++col) {
        Table.Column column = table.rows().get(row).get(col);
```
is repeated in four methods. Maybe we can have a method in `Table` to provide this iteration

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderReferenceCheck.java
```diff
@@ -0,0 +1,398 @@
...
+      for (int col = 0; col < table.rows().get(row).size(); ++col) {
+        Table.Column column = table.rows().get(row).get(col);
+        if (column != null) {
+          TagNode node = column.node();
+          if (hasHeaders(node)) {
+            List<String> actual = getHeaders(column.node());
+            List<String> expected = referenceableHeaders.getOrDefault(node, new ArrayList<>());
+            for (String header : actual) {
```
saberduck: I think you can extract this verification into separate method

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderReferenceCheck.java
```diff
@@ -0,0 +1,398 @@
...
+  }
+
+  @Override
+  public void endElement(TagNode node) {
+    if (isTable(node) && !stack.isEmpty()) {
+      raiseViolationOnInvalidReference(stack.pop().build());
+    }
+  }
+
+  private void raiseViolationOnInvalidReference(Table table) {
```
saberduck: You can move all methods which have just `Table` as parameter to the table class, so they can be instance methods and access the fields implicitly.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderReferenceCheck.java
```diff
@@ -0,0 +1,398 @@
...
+                break;
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+
+  private static List<Set<String>> findHorizontalHeaders(Table table) {
```
saberduck: same - can be moved to `Table` class

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderReferenceCheck.java
```diff
@@ -0,0 +1,398 @@
...
+        if (column instanceof Table.Header) {
+          Table.Header header = (Table.Header) column;
+          headers.get(col).add(header.id());
+        }
+      }
+    }
+    return headers;
+  }
+
+  private static List<Set<String>> findVerticalHeaders(Table table) {
```
saberduck: same - can be moved to `Table` class

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderReferenceCheck.java
```diff
@@ -0,0 +1,398 @@
...
+        if (column instanceof Table.Header) {
+          Table.Header header = (Table.Header) column;
+          headers.get(row).add(header.id());
+        }
+      }
+    }
+    return headers;
+  }
+
+  private static Map<TagNode, List<String>> findReferenceableHeadersPerColumnNode(Table table) {
```
saberduck: same - can be moved to `Table` class

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderReferenceCheck.java
```diff
@@ -0,0 +1,398 @@
...
+      for (int row = rowStart; row < rowEnd; ++row) {
+        if (row == rows.size()) {
+          rows.add(new RowBuilder());
+        }
+        for (int col = columnStart; col < columnEnd; ++col) {
+          if (col < rows.get(row).size()) {
+            rows.get(row).set(col, column);
+          } else {
+            for (int i = rows.get(row).size(); i < col; ++i) {
+              rows.get(row).add(null);
```
saberduck: I would avoid using `null` as value in collection, you can perhaps create an instance of `Column` class to represent this value

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderReferenceCheck.java
```diff
@@ -0,0 +1,398 @@
...
+  }
+
+  private void raiseViolationOnInvalidReference(Table table) {
+    Map<TagNode, List<String>> referenceableHeaders = findReferenceableHeadersPerColumnNode(table);
+    for (int row = 0; row < table.rows().size(); ++row) {
+      for (int col = 0; col < table.rows().get(row).size(); ++col) {
+        Table.Column column = table.rows().get(row).get(col);
+        if (column != null) {
+          TagNode node = column.node();
+          if (hasHeaders(node)) {
```
saberduck: maybe `column` can provide `headers()` method, which can return empty list when there are none. This would simplify a bit logic here.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderReferenceCheck.java
```diff
@@ -0,0 +1,398 @@
...
+    return referenceable;
+  }
+
+  private static boolean isExistingHeader(Table table, String headerName) {
+    return table.rows().stream().flatMap(List::stream).filter(column -> column instanceof Table.Header)
+        .map(column -> (Table.Header) column).anyMatch(header -> headerName.equalsIgnoreCase(header.id()));
+  }
+
+  private static boolean isTable(TagNode node) {
+    return "TABLE".equalsIgnoreCase(node.getNodeName());
```
saberduck: use `TagNode#equalsElementName`

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderReferenceCheck.java
```diff
@@ -0,0 +1,398 @@
...
+  }
+
+  private static boolean isTableRow(TagNode node) {
+    return "TR".equalsIgnoreCase(node.getNodeName());
```
saberduck: use `TagNode#equalsElementName`

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderReferenceCheck.java
```diff
@@ -0,0 +1,398 @@
...
+  }
+
+  private static boolean isTableColumn(TagNode node) {
+    return "TD".equalsIgnoreCase(node.getNodeName());
```
saberduck: use `TagNode#equalsElementName`

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderReferenceCheck.java
```diff
@@ -0,0 +1,398 @@
...
+
+  private static boolean isTable(TagNode node) {
+    return "TABLE".equalsIgnoreCase(node.getNodeName());
+  }
+
+  private static boolean isTableRow(TagNode node) {
+    return "TR".equalsIgnoreCase(node.getNodeName());
+  }
+
+  private static boolean isTableColumn(TagNode node) {
```
saberduck: method name is not ideal, this is table data, not column

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderReferenceCheck.java
```diff
@@ -0,0 +1,398 @@
...
+  private static boolean isTableHeader(TagNode node) {
+    return "TH".equalsIgnoreCase(node.getNodeName());
+  }
+
+  private static boolean hasHeaders(TagNode node) {
+    return node.hasProperty(HEADERS) && !isDynamicHeaders(node);
+  }
+
+  private static boolean isDynamicHeaders(TagNode node) {
+    return DYNAMIC_HEADERS.matcher(node.getPropertyValue(HEADERS)).find();
```
saberduck: you can inline this method, it's used only from `hasHeaders` method

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderReferenceCheck.java
```diff
@@ -0,0 +1,398 @@
...
+  }
+
+  private static List<String> getHeaders(TagNode node) {
+    return Arrays.asList(node.getPropertyValue(HEADERS).split("\\s+")).stream()
+        .filter(header -> !header.isEmpty()).collect(Collectors.toList());
+  }
+
+  private static boolean hasRowSpan(TagNode node) {
```
saberduck: I find these methods redundant, they have no logic, just replacing instance method with static one, you can avoid using them by using `getRowSpan` and providing default value when attribute is not present

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderReferenceCheck.java
```diff
@@ -0,0 +1,398 @@
...
+    return node.hasProperty("ROWSPAN");
+  }
+
+  private static boolean hasColSpan(TagNode node) {
```
saberduck: same as above for `hasRowSpan`

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderReferenceCheck.java
```diff
@@ -0,0 +1,398 @@
...
+      }
+
+      public void add(@Nullable Table.Column column) {
+        columns.add(column);
+      }
+    }
+
+    public void newRow() {
+      rows.add(new RowBuilder());
+      currentRow = rows.get(rows.indexOf(currentRow) + 1);
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: Not necessarily. When we are dealing with columns that have a rowspan of n > 2, we will potentially add n new rows. For that reason, when `newRow` is invoked, the current row will be the first one among those that come after the current `currentRow`.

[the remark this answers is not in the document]
yassin-kammoun-sonarsource: I notice now that I don't check if there is already a new row available instead of adding one systematically. I will change that.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderReferenceCheck.java
```diff
@@ -0,0 +1,398 @@
...
+    }
+
+    public void newColumn(Table.Column column) {
+      if (!rows.isEmpty()) {
+        TagNode node = column.node();
+        if (hasColSpan(node) && hasRowSpan(node)) {
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: Indeed. `newColumnWithRowColSpan` was actually the generic method you are describing.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderReferenceCheck.java
yassin-kammoun-sonarsource: `Table#forEachCell` reduces the implementation so there is probably no need for a separate method now.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderReferenceCheck.java
yassin-kammoun-sonarsource: We can do that for all methods except this one; we cannot invoke `createViolation` within `Table` since this class is static.

saberduck: indeed, makes sense

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderReferenceCheck.java
```diff
@@ -0,0 +1,335 @@
...
+        this.node = node;
+      }
+
+      TagNode node() {
+        return node;
+      }
+
+      List<String> headers() {
+        if (node.hasProperty(HEADERS) && !DYNAMIC_HEADERS.matcher(node.getPropertyValue(HEADERS)).find()) {
+          return Arrays.asList(node.getPropertyValue(HEADERS).split("\\s+")).stream().filter(header -> !header.isEmpty()).collect(Collectors.toList());
```
saberduck: You can use `Arrays.stream` instead

PR #137, opened 2019-08-15
SONARHTML-111 Support precise issue location

saberduck:


sonar-html-plugin/src/main/java/org/sonar/plugins/html/core/HtmlSensor.java
```diff
@@ -133,18 +133,25 @@ private static void saveMetrics(SensorContext context, HtmlSourceCode sourceCode
...
     }
   }
 
+  private static NewIssueLocation locationForIssue(InputFile inputFile, HtmlIssue issue, NewIssue newIssue) {
+    NewIssueLocation location = newIssue.newLocation()
+      .on(inputFile)
+      .message(issue.message());
+    Integer line = issue.line();
+    if (issue.startColumn() != null) {
+      location.at(inputFile.newRange(issue.line(), issue.startColumn(), issue.endLine(), issue.endColumn()));
```
quentin-jaquier-sonarsource: Do you assume that if startColumn is not null, endLine and endColumn will not be either? Or is it fine to have two last as null? 

If it's the first case I think we should change it!

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/FieldsetWithoutLegendCheck.java
```diff
@@ -40,12 +40,12 @@ public void startElement(TagNode node) {
   @Override
   public void endElement(TagNode node) {
     if (isFieldSet(node)) {
-      if (!foundLegend && fieldsetLine != 0) {
-        createViolation(fieldsetLine, "Add a <legend> tag to this fieldset.");
+      if (!foundLegend && fieldset != null) {
```
quentin-jaquier-sonarsource: Is it possible that node.getStartLinePosition() returns 0?
If no, everything is fine, if yes we miss the case where fieldset is not null but equals to zero!

The same happen in FlashUsesBothObjectAndEmbedCheck.java.

saberduck: theoretically, you are right, but the rule doesn't care about the special case of fieldset at line 0, it should detect the fieldset regardless of its position.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/HtmlIssue.java
```diff
@@ -60,4 +68,16 @@ public Double cost() {
     return cost;
   }
 
+  public Integer endLine() {
```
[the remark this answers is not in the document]
saberduck: Indeed, this was not very nice, I handled it differently, with a subclass which doesn't use null values

sonar-html-plugin/src/main/java/org/sonar/plugins/html/core/HtmlSensor.java
saberduck: indeed, this different cases didn't make much sense, I think it's nicer with a specialized subclass for precise issue location

sonar-html-plugin/src/main/java/org/sonar/plugins/html/core/HtmlSensor.java
```diff
@@ -133,18 +134,29 @@ private static void saveMetrics(SensorContext context, HtmlSourceCode sourceCode
...
       newIssue.save();
     }
   }
 
+  private static NewIssueLocation locationForIssue(InputFile inputFile, HtmlIssue issue, NewIssue newIssue) {
+    NewIssueLocation location = newIssue.newLocation()
+      .on(inputFile)
+      .message(issue.message());
+    Integer line = issue.line();
+    if (issue instanceof PreciseHtmlIssue) {
```
quentin-jaquier-sonarsource: 👍

PR #152, opened 2020-11-06
SONARHTML-116 Rule S1935 is raising on "<style>" elements instead of "style" attributes

quentin-jaquier-sonarsource:


margarita-nedzelska-sonarsource (approved):
Looks good!

PR #153, opened 2020-11-06
SONARHTML-115 FP from Rule S1085 when "aria-label" or "aria-labelledby" are used

quentin-jaquier-sonarsource:


margarita-nedzelska-sonarsource (approved):
LGTM 👍

PR #159, opened 2020-11-09
SONARHTML-127 Rule S5148: update rule type to Security Hotspot and accept using "noopener" without "noreferrer"

margarita-nedzelska-sonarsource:


quentin-jaquier-sonarsource (approved):
LGTM

PR #160, opened 2020-11-09
Fix quality flaw:

margarita-nedzelska-sonarsource:
reduce visibility of abstract class constructor
  add missing package.info
  add missing locale
  add missing Nullable


quentin-jaquier-sonarsource (approved):
LGTM

PR #163, opened 2020-11-10
SONARHTML-120 Enable support for SECURITY_HOTSPOT

quentin-jaquier-sonarsource:
I first found out that if you are not providing a `SonarRuntime` to `RuleMetadataLoader`, [Security Hotspots Rules will be considered as Vulnerability rules](https://github.com/SonarSource/sonar-analyzer-commons/commit/edb7dc60689a2ef28b8d283241fd6eefe7698ed4#diff-485ea7f7350a405f85a7985bac360f4b7a6c3787a674dbe9a0b556dd0b2add65R67-R77). 

Then, I realize that updating the "analyzer commons" version to the latest also solves the problem, in a cleaner way.
It makes sense since this plugin will always target the latest version of SonarQube, definitely supporting security hotspots.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -66,7 +66,7 @@
     <junit.version>4.12</junit.version>
     <sslr.version>1.22</sslr.version>
     <assertj.version>3.17.2</assertj.version>
-    <analyzerCommons.version>1.10.3.509</analyzerCommons.version>
+    <analyzerCommons.version>1.12.0.632</analyzerCommons.version>
 
     <artifactsToPublish>${project.groupId}:sonar-html-plugin:jar</artifactsToPublish>
     <maven.compiler.release>8</maven.compiler.release>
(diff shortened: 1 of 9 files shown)


pom.xml
```diff
@@ -66,7 +66,7 @@
     <junit.version>4.12</junit.version>
     <sslr.version>1.22</sslr.version>
     <assertj.version>3.17.2</assertj.version>
-    <analyzerCommons.version>1.10.3.509</analyzerCommons.version>
+    <analyzerCommons.version>1.12.0.632</analyzerCommons.version>
```
m-g-sonar, line 69: this makes me think that on pl-sql... we didn't changed the version and we are still relying on 1.9. Does it mean that hotspots are not working?!

quentin-jaquier-sonarsource: I had the same reaction, and checked PL/SQL: good news, everything is working as expected, as we are providing the `SonarRuntime` to `RuleMetadataLoader`.

PR #167, opened 2020-11-11
SONARHTML-126 Update rules metadata

margarita-nedzelska-sonarsource:


sonarpedia.json
```diff
@@ -1,11 +1,11 @@
 {
   "rules-metadata-path": "./sonar-html-plugin/src/main/resources/org/sonar/l10n/web/rules/Web",
   "languages": [
-    "WEB"
```
quentin-jaquier-sonarsource, line 4: I think we should keep WEB, it's working fine and we avoid unexpected side-effects. In addition, we should not bother too much with rule-api magic since it will (hopefully) be replaced soon.

PR #171, opened 2021-01-16
SONARHTML-131 Remove dependency on sslr-squid-bridge because it's deprecated

margarita-nedzelska-sonarsource:


pavel-mikula-sonarsource (approved):
LGTM
There are some comments about squid bridge, should they be removed as well?

https://github.com/SonarSource/sonar-html/blob/ae3915ccf652b21d1de243e07943c22820857721/sonar-html-plugin/src/test/java/org/sonar/plugins/html/checks/CheckMessagesVerifier.java#L40

https://github.com/SonarSource/sonar-html/blob/ae3915ccf652b21d1de243e07943c22820857721/sonar-html-plugin/src/test/java/org/sonar/plugins/html/checks/CheckMessagesVerifierRule.java#L39

margarita-nedzelska-sonarsource:
Thanks, for these findings. I've tried to understand whether they're useful and I guess it will cause no harm if I remove these lines :)

PR #185, opened 2021-11-15
SONARHTML-144 Support SalesForce Aura Lightning Components: update default suffixes with ".cmp"

yassin-kammoun-sonarsource:


its/plugin/pom.xml
```diff
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.sonarsource.html</groupId>
     <artifactId>html-its</artifactId>
-    <version>3.4.0-SNAPSHOT</version>
+    <version>3.5.0-SNAPSHOT</version>
```
roberto-orlandi-sonarsource, line 8: was it intended to increase also the plugin version in this PR?

yassin-kammoun-sonarsource: Oups, I will revert the commit.

roberto-orlandi-sonarsource: I will wait for it to be green again and then I'll approve it :)

nils-werner-sonarsource: However, I just noticed, that the its/plugin pom version is not identical to the other pom versions. Should be fixed in another commit.

yassin-kammoun-sonarsource: Thanks, I will create a separate PR for this.

roberto-orlandi-sonarsource: It seems that it was decreased few minutes ago with this commit:
https://github.com/SonarSource/sonar-html/commit/0f76403603148d7a50037436b05bede658e24ba2

roberto-orlandi-sonarsource: Forget my last comment :D

roberto-orlandi-sonarsource (approved):
LGTM!

PR #197, opened 2022-01-04
SONARHTML-150 Declare HTML sensor as a sensor processing files independently

mpaladin:


sonar-html-plugin/src/test/java/org/sonar/plugins/html/core/HtmlSensorTest.java
```diff
@@ -155,6 +165,41 @@ public void testDescriptor() {
...
+    DefaultSensorDescriptor sensorDescriptor = new DefaultSensorDescriptor();
+    SonarRuntime sonarRuntime = SonarRuntimeImpl.forSonarLint(Version.create(6, 5));
+    new HtmlSensor(sonarRuntime, null, null, new CheckFactory(new DefaultActiveRules(Collections.emptyList()))).describe(sensorDescriptor);
+    assertThat(sensorDescriptor.name()).isEqualTo("HTML");
+    assertThat(sensorDescriptor.languages()).isEmpty();
+  }
+
+  @Test
+  public void test_descriptor_sonarqube_9_3() {
+    final boolean[] called = {false};
```
malte-skoruppa-sonarsource, line 179: A creative way of instantiating a mutable boolean! :grin: 

Just as a side note, there exists the class `java.util.concurrent.atomic.AtomicBoolean` that wraps a boolean value that can be updtaed atomically that you could use here as well. But an array of size 1 works just as good, and anyway this is just test code and it's admittedly a matter of taste! :smile:

malte-skoruppa-sonarsource (approved):
LGTM!

PR #201, opened 2022-01-18
SONARHTML-140 S1077 should not raise an issue on "<img th:alt=..."

karim-ouerghemmi-sonarsource:
Theoretically, these Thymeleaf attributes would need to be considered in many checks in which we look for the absence or at the value of an attribute ([doc](https://www.thymeleaf.org/doc/tutorials/3.0/usingthymeleaf.html#setting-value-to-specific-attributes)). 
However, I wanted to avoid going down a rabbit hole, and only fixed the specific FP of the ticket (`alt` attribute of an `img` tag).


roberto-orlandi-sonarsource (approved):
LGTM

PR #202, opened 2022-01-19
SONARHTML-147 S103 should not raise an issue on <svg><path d=xxx> property

karim-ouerghemmi-sonarsource:


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/coding/MaxLineLengthCheck.java
```diff
@@ -39,20 +42,38 @@
     defaultValue = "" + DEFAULT_MAX_LINE_LENGTH)
   public int maxLength = DEFAULT_MAX_LINE_LENGTH;
 
+  private final Set<Integer> ignoredLines = new HashSet<>();
+
   @Override
   public void startDocument(List<Node> nodes) {
+    ignoredLines.clear();
```
roberto-orlandi-sonarsource, line 49: would it work as well if we get rid of overriding `startDocument `and we add this instruction at the end of `endDocument`?

roberto-orlandi-sonarsource (approved):
LGTM, I left just one comment that you may want or not consider :)

PR #203, opened 2022-01-19
SONARHTML-151 Web:TableHeaderHasIdOrScopeCheck should not raise for s…

roberto-orlandi-sonarsource:
…imple tables


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderHasIdOrScopeCheck.java
```diff
@@ -17,22 +17,100 @@
...
+      // Sometimes rows are defined in separate files, and at this point we don't see them as part of a current table.
+      // In this case we treat them as belonging to "not simple tables".
+      raiseIssueOnTableHeadersWithoutScopeOrId(row);
+    }
+  }
+
+  @Override
+  public void endElement(TagNode node) {
+    if (isTableTag(node) && !tables.isEmpty()) {
+      TableElement currentTable = tables.peek();
```
karim-ouerghemmi-sonarsource, line 73: _minor_ I think we can directly `pop()` here instead of `peek()` and remove line `77`

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderHasIdOrScopeCheck.java
```diff
@@ -17,22 +17,100 @@
...
 import org.sonar.check.Rule;
 import org.sonar.plugins.html.checks.AbstractPageCheck;
 import org.sonar.plugins.html.node.TagNode;
 
 @Rule(key = "TableHeaderHasIdOrScopeCheck")
 public class TableHeaderHasIdOrScopeCheck extends AbstractPageCheck {
 
+  private static final String MESSAGE = "Add either an 'id' or a 'scope' attribute to this <th> tag.";
+
+  private Deque<TableElement> tables = new ArrayDeque<>();
```
karim-ouerghemmi-sonarsource, line 34: _minor_ we can make this field `final`

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderHasIdOrScopeCheck.java
```diff
@@ -17,22 +17,100 @@
...
+  @Override
+  public void endElement(TagNode node) {
+    if (isTableTag(node) && !tables.isEmpty()) {
+      TableElement currentTable = tables.peek();
+      if (!currentTable.isSimpleTable()) {
+        raiseIssueOnTableHeadersWithoutScopeOrId(currentTable.headers);
+      }
+      tables.pop();
+    }
   }
```
karim-ouerghemmi-sonarsource, line 80: As a defensive measure, I would override `endDocument()` and put a `tables.clear()` in there. 

As it is the same check instance that gets used for every file, this would avoid cross-file state contamination (<- sounds like a vulnerability :laughing:) when a table does not get closed.
For example, if the first file being scanned is
```html
<table>
    <tr>
        <th>foo</th>
        <th>bar</th>
    </tr>
``` 
and the second is:
```html
<tr>
    <th>foo</th>
    <th>bar</th>
</tr>
</table>
```
This could lead to behavior that might be not so nice to debug in the future. Also memory-wise we would avoid the risk of keeping references to the nodes of a previous file.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderHasIdOrScopeCheck.java
```diff
@@ -17,22 +17,100 @@
...
+  }
+
+  private static class TableElement {
+    List<TagNode> headers = new ArrayList<>();
+    List<TagNode> firstRow = new ArrayList<>();
+    List<TagNode> firstCol = new ArrayList<>();
+
+    /**
+     * We consider as simple tables, tables which have all headers only in the first row or in the first column.
+     * If both first row and first column are all (or partially) composed by headers are not considered simple tables.
```
karim-ouerghemmi-sonarsource, line 110: I'm not sure I understand the second sentence. Did you mean to say something like "If both first row and first column are all (or partially) composed by headers, the table is not considered a simple tables."

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderHasIdOrScopeCheck.java
```diff
@@ -17,22 +17,100 @@
...
+     **/
+    boolean isSimpleTable() {
+      return headers.equals(firstRow) || headers.equals(firstCol);
```
karim-ouerghemmi-sonarsource, line 113: Nice idea :+1:

karim-ouerghemmi-sonarsource (approved):
LGTM. Nice solution!
Left some mostly _minor_ comments.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderHasIdOrScopeCheck.java
roberto-orlandi-sonarsource: Indeed, good point, thank you!

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderHasIdOrScopeCheck.java
roberto-orlandi-sonarsource: yes, but since it is not clear I will update this. Does "We consider as simple tables, tables with the first row or the first column only composed by headers.  The two conditions must not apply together." sounds clearer?

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderHasIdOrScopeCheck.java
roberto-orlandi-sonarsource: > Nice idea 👍

Thanks, it's a last moment one, replacing several complex boolean expressions candidated to represent a simple table :D

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderHasIdOrScopeCheck.java
roberto-orlandi-sonarsource: Wait, this is ambiguous.. I will reformulate :D

roberto-orlandi-sonarsource: "Simple tables are considered as such when the headers are either all in the first row, or all in the first column. The two conditions must not apply together."

PR #204, opened 2022-01-20
SONARHTML-141 Add the `.twig` file extension to the default html suffix list

karim-ouerghemmi-sonarsource:


roberto-orlandi-sonarsource (approved):
LGTM!

PR #206, opened 2022-01-24
Update GitHub actions release flow to v4

roberto-orlandi-sonarsource:


karim-ouerghemmi-sonarsource (approved):
Not sure if strictly necessary to fix the problem, but to be safe let's also update to `v4` the `download-build` and `maven-central-sync` actions. See example commit on Slang https://github.com/SonarSource/slang/commit/65a128adc369cc040db28533cf8216a526b0be08

PR #207, opened 2022-01-24
Prepare for next development iteration

roberto-orlandi-sonarsource:


karim-ouerghemmi-sonarsource (approved):
LGTM, but seems that `git add` added one file too much: `rule-api-2.2.0.1998.jar` :laughing:

PR #220, opened 2022-11-22
SONARHTML-138 S5254 language_attributes() call should be a valid language attribute supplier

thomas-serre-sonarsource:


sonar-html-plugin/src/test/resources/checks/LangAttributeCheck.html
```diff
@@ -13,3 +13,5 @@
 <html lang="en" xml:lang="en"> <!-- Compliant -->
   <head></head>
 </html>
+
+<html  <?php language_attributes(); ?>> <head></head> </html> <!-- Compliant -->
```
christophe-zurn-sonarsource, line 17: _minor_ this is an internal convention to end files with newlines

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/LangAttributeCheck.java
```diff
@@ -36,6 +37,21 @@ private static boolean isHtmlTag(TagNode node) {
...
+   * Using WordPress, HTML attributes can be set using the php function `language_attributes`
+   */
+  private static boolean hasLangAttributeFromWordPress(TagNode node) {
+    for (Attribute attribute : node.getAttributes()) {
+      String attributeName = attribute.getName();
+      if (attributeName.contains("?php") && attributeName.contains("language_attributes")) {
+        return true;
+      }
+    }
+    return false;
```
christophe-zurn-sonarsource, line 55: This is a matter of preference of implementation, but it's also possible to use Java 8 streams for this logic as follows:
```suggestion
    return node.getAttributes().stream()
      .map(Attribute::getName)
      .anyMatch(attributeName -> attributeName.contains("?php") && attributeName.contains("language_attributes"));
```
Up to you to keep as is or update :)

christophe-zurn-sonarsource (approved):
LGTM !
Nicely done :)

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/LangAttributeCheck.java
thomas-serre-sonarsource: I chose the stream version ;) Thanks for the java refresher :)

PR #221, opened 2022-11-22
SONARHTML-139 Should not raise issue on comments annotated with @elvariable or @thymesVar

thomas-serre-sonarsource:


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/comments/AvoidCommentedOutCodeCheck.java
```diff
@@ -40,6 +41,8 @@ public class AvoidCommentedOutCodeCheck extends AbstractPageCheck {
 
   private static final CodeRecognizer CODE_RECOGNIZER = new CodeRecognizer(THRESHOLD, LANGUAGE_FOOTPRINT);
 
+  private static final List<String> IGNORED_COMMENT_ANNOTATIONS = Arrays.asList("@thymesVar", "@elvariable");
```
christophe-zurn-sonarsource, line 44: _minor_ since java 9, we usually use `List.of` when creating static final lists in classes, as it will return an immutable list.

christophe-zurn-sonarsource (approved):
LGTM !
I left a small comment regarding the list creation. I let you see whether you want to act on it or not :)

PR #222, opened 2022-11-24
SONARHTML-160 S5254 should not raise an issue when the lang is specified with Thymeleaf attributes

thomas-serre-sonarsource:


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/LangAttributeCheck.java
```diff
@@ -39,15 +39,31 @@ private static boolean isHtmlTag(TagNode node) {
...
       || node.hasProperty("xml:lang")
-      || hasLangAttributeFromWordPress(node);
+      || hasWordPressLangAttribute(node)
+      || hasThymeleafLangAttribute(node);
   }
 
   /**
    * Using WordPress, HTML attributes can be set using the php function `language_attributes`
    */
-  private static boolean hasLangAttributeFromWordPress(TagNode node) {
```
thomas-serre-sonarsource, line 48: Renaming to match function name from [SONARHTML-140](https://sonarsource.atlassian.net/browse/SONARHTML-140)

sonar-html-plugin/src/test/resources/checks/LangAttributeCheck.html
```diff
@@ -15,3 +15,23 @@
 </html>
 
 <html  <?php language_attributes(); ?>> <head></head> </html> <!-- Compliant -->
+
+<html th:lang="${#locale.language}" > <!-- Compliant -->
+<head></head>
+</html>
```
christophe-zurn-sonarsource, line 21: This sample (and the others) could be reduced a bit to simplify the example:
```suggestion
<html th:lang="${#locale.language}"></html> <!-- Compliant -->
```

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/LangAttributeCheck.java
```diff
@@ -39,15 +39,31 @@ private static boolean isHtmlTag(TagNode node) {
...
+   * In Thymeleaf there are multiple ways of specifying the lang attribute:
+   * - using the th:lang, th:xmllang, th:lang-xmllang attributes (lang-xmllang would set both xmllang and lang attributes)
+   * - using the th:attr attribute for specifying different attributes. Example "th:attr="lang=html,xml:lang=html""
+   */
+  private static boolean hasThymeleafLangAttribute(TagNode node) {
+    String thAttrValue = node.getAttribute("th:attr");
+    return node.hasProperty("th:lang")
+      || node.hasProperty("th:xmllang")
+      || node.hasProperty("th:lang-xmllang")
+      || (thAttrValue != null && thAttrValue.contains("lang")
```
christophe-zurn-sonarsource, line 65: To reduce the possibility of False Negatives, it might be nicer to check for `thAttrValue.contains("lang=")` instead, otherwise if the `lang` string appears anywhere in the attribute, it will be considered valid. Example: `<html th:attr="class="my-lang-class"" >` would be wrongly considered as compliant.
False Negatives would still be possible with the check for `lang=`, (for instance in the case of `<html th:attr="title="lang=en"" >`.) but would be less likely.

sonar-html-plugin/src/test/resources/checks/LangAttributeCheck.html
```diff
@@ -15,3 +15,23 @@
...
+
+<html th:attr="lang="en"" > <!-- Compliant -->
+<head></head>
+</html>
+
+<html th:attr="xml:lang="en"" > <!-- Compliant -->
+<head></head>
+</html>
```
christophe-zurn-sonarsource, line 37: I think it's worth adding the False Negative code examples mentioned before here as well, just to document them :)

christophe-zurn-sonarsource (changes requested):

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/LangAttributeCheck.java
thomas-serre-sonarsource: Sounds good !

sonar-html-plugin/src/test/resources/checks/LangAttributeCheck.html
thomas-serre-sonarsource: I add one False positive case. Let me know if you think about other potential false positives.

PR #223, opened 2022-11-24
SONARHTML-162 The '<' character can incorrectly be interpreted as a start tag

christophe-zurn-sonarsource:


sonar-html-plugin/src/main/java/org/sonar/plugins/html/lex/NormalElementTokenizer.java
```diff
@@ -33,9 +33,6 @@ public boolean consume(CodeReader codeReader, List<Node> nodeList) {
       return false;
     }
     int i = 1;
-    while (Character.isWhitespace(codeReader.charAt(i))) {
-      i++;
-    }
```
thomas-serre-sonarsource, line 38: This code is useless because it is incorrect in HTML to have something like ```<     html>``` compared to ```<html>```, right?

sonar-html-plugin/src/test/java/org/sonar/plugins/html/lex/PageLexerTest.java
```diff
@@ -503,8 +502,32 @@ public void cdata() {
...
+    assertThat(nodes).extracting(Node::getCode).containsExactly("<a>", " < a ", "</a>");
+  }
+
+  @Test
+  public void start_tag_character_is_considered_as_text_when_last_character_of_code() {
+    PageLexer lexer = new PageLexer();
+    List<Node> nodes = lexer.parse(new StringReader("<a> <"));
+    assertThat(nodes).hasSize(2);
+    assertThat(nodes).extracting(Node::getNodeType).containsExactly(NodeType.TAG, NodeType.TEXT);
+    assertThat(nodes).extracting(Node::getCode).containsExactly("<a>", " <");
```
thomas-serre-sonarsource, line 530: From my understanding, with this test, you are verifying that 
```java
return followingChar == 0 && ...
```
In the function, `isFollowedByWhitespaceCharacter` is working, right?

thomas-serre-sonarsource (approved):

thomas-serre-sonarsource (changes requested):
see https://github.com/SonarSource/sonar-html/pull/223/files#r1032418861

sonar-html-plugin/src/main/java/org/sonar/plugins/html/lex/NormalElementTokenizer.java
```diff
@@ -33,16 +33,51 @@ public boolean consume(CodeReader codeReader, List<Node> nodeList) {
...
+      || (currentChar >= '\u00F8' && currentChar <= '\u02FF')
+      || (currentChar >= '\u0370' && currentChar <= '\u037D')
+      || (currentChar >= '\u037F' && currentChar <= '\u1FFF')
+      || (currentChar >= '\u200C' && currentChar <= '\u200D')
+      || (currentChar >= '\u2070' && currentChar <= '\u218F')
+      || (currentChar >= '\u2C00' && currentChar <= '\u2FEF')
+      || (currentChar >= '\u3001' && currentChar <= '\uD7FF')
+      || (currentChar >= '\uF900' && currentChar <= '\uFDCF')
+      || (currentChar >= '\uFDF0' && currentChar <= '\uFFFD');
+  }
```
thomas-serre-sonarsource, line 71: _minor_ My Python 🐍 corrupted mind would prefer a construct like the one below, but it is not elegant as in Python ;). Anyway, I'm okay with your code.

```suggestion
  static final ImmutableRangeSet<Character> valid characters =  new ImmutableRangeSet.Builder<Character>()
    .add(Range.singleton(':'))
    .add(Range.closed('A', 'Z'))
    .add(Range.singleton('_'))
    .add(Range.closed('a', 'z'))
    .add(Range.closed('\u00C0','\u00D6'))
    .add(Range.closed('\u00D8','\u00F6'))
    .add(Range.closed('\u00F8','\u02FF'))
    .add(Range.closed('\u0370','\u037D'))
    .add(Range.closed('\u037F','\u1FFF'))
    .add(Range.closed('\u200C','\u200D'))
    .add(Range.closed('\u2070','\u218F'))
    .add(Range.closed('\u2C00','\u2FEF'))
    .add(Range.closed('\u3001','\uD7FF'))
    .add(Range.closed('\uF900','\uFDCF'))
    .add(Range.closed('\uFDF0','\uFFFD'))
    .build();
  static boolean isValidSingleCharCodeNameStartChar(char currentChar) {
    return valid characters.contains(currentChar);
  }
```

sonar-html-plugin/src/main/java/org/sonar/plugins/html/lex/NormalElementTokenizer.java
```diff
@@ -33,16 +33,51 @@ public boolean consume(CodeReader codeReader, List<Node> nodeList) {
...
+
+  // Visible for testing
+  static boolean isValidSurrogatePairNameStartChar(char currentChar, char nextChar) {
+    if (!Character.isSurrogatePair(currentChar, nextChar)) {
+      return false;
     }
-    return i > elementNameIndex && super.consume(codeReader, nodeList);
+
+    int codePoint = Character.toCodePoint(currentChar, nextChar);
+    return codePoint >= 0x10000 && codePoint <= 0xEFFFF;
```
thomas-serre-sonarsource, line 80: _curiosity_ I understand that here you are matching for the last possible pattern of 'NameStartChar', cf. [this](https://www.w3.org/TR/REC-xml/#NT-NameStartChar). 

How did you know that the pattern ```[#x10000-#xEFFFF]``` is not simply the concatenation of the four bytes of the two characters but rather must be computed using ```Character.toCodePoint()```?

sonar-html-plugin/src/test/java/org/sonar/plugins/html/lex/PageLexerTest.java
```diff
@@ -503,8 +502,34 @@ public void cdata() {
...
+    assertOnlyText("<\u2190html>");
+  }
+
+  @Test
+  public void start_tag_character_is_considered_as_text_when_followed_by_whitespace() {
+    PageLexer lexer = new PageLexer();
+    List<Node> nodes = lexer.parse(new StringReader("<a> < a </a>"));
+    assertThat(nodes).hasSize(4);
+    assertThat(nodes).extracting(Node::getNodeType).containsExactly(NodeType.TAG, NodeType.TEXT, NodeType.TEXT, NodeType.TAG);
+    assertThat(nodes).extracting(Node::getCode).containsExactly("<a>", " ", "< a ", "</a>");
```
thomas-serre-sonarsource, line 523: _curiosity_ Do you see any potential issue having two or more consecutive Text nodes? Shouldn't we merge them ?

thomas-serre-sonarsource (approved):
Impressive work! It looks nice to me. I left a few minor comments and questions out of curiosity.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/lex/NormalElementTokenizer.java
christophe-zurn-sonarsource: I think your version is actually more readable.
The thing is that `ImmutableRangeSet` is from the `guava` library, and we try to avoid it as much as possible adding dependencies to keep the analyzers as lightweight as possible.
Another suggestion, to improve readability without adding a new dependency: how about having an additional `isInCharRange` helper method? That way the code would look like:
```
  static boolean isValidSingleCharCodeNameStartChar(char currentChar) {
    return currentChar == ':'
      || currentChar == '_'
      || isInCharRange(currentChar, 'A', 'Z')
      || isInCharRange(currentChar, 'a', 'z')
      || isInCharRange(currentChar, '\u00C0', '\u00D6')
      || isInCharRange(currentChar, '\u00D8', '\u00F6')
      || isInCharRange(currentChar, '\u00F8', '\u02FF')
      || isInCharRange(currentChar, '\u0370', '\u037D')
      || isInCharRange(currentChar, '\u037F', '\u1FFF')
      || isInCharRange(currentChar, '\u200C', '\u200D')
      || isInCharRange(currentChar, '\u2070', '\u218F')
      || isInCharRange(currentChar, '\u2C00', '\u2FEF')
      || isInCharRange(currentChar, '\u3001', '\uD7FF')
      || isInCharRange(currentChar, '\uF900', '\uFDCF')
      || isInCharRange(currentChar, '\uFDF0', '\uFFFD');
  }
```

sonar-html-plugin/src/test/java/org/sonar/plugins/html/lex/PageLexerTest.java
christophe-zurn-sonarsource: I thought about it, but merging them was not so simple to do easily.
I've checked, and the only checks using text nodes are `LongJavaScriptCheck`, `FileLengthCheck`, `LinksIdenticalTextsDifferentTargetsCheck`, `ObjectWithAlternativeContentCheck`, and `InternationalizationCheck`. For the first four, having consecutive text nodes is not a problem, as the text is either appended in the rule, or the check itself makes it so it does not matter.
For `InternationalizationCheck`, it could potentially lead to a FP in the case there is a '<' character in a text node and the characters before it are only letters or digits. Ex: `<span>r< </span>`

sonar-html-plugin/src/main/java/org/sonar/plugins/html/lex/NormalElementTokenizer.java
christophe-zurn-sonarsource: In the `CodeReader` we transform the initial input stream (from its original format `UTF-8`, `UTF-16`, `UTF-32`, ...) into a UTF-16 char array (which is the Java representation of Unicode)

sonar-html-plugin/src/main/java/org/sonar/plugins/html/lex/NormalElementTokenizer.java
thomas-serre-sonarsource: Understandable, alright. I prefer the first one at the end.

christophe-zurn-sonarsource: Done: I've updated the code back to the original, more readable, version

PR #226, opened 2022-11-28
SONARHTML-155 Web:UnifiedExpressionCheck should not raise on thymeleaf attributes

thomas-serre-sonarsource:


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/scripting/UnifiedExpressionCheck.java
```diff
@@ -122,6 +122,13 @@ public void startDocument(List<Node> nodes) {
   public void startElement(TagNode element) {
 
     for (Attribute attribute : element.getAttributes()) {
+
+      // Ignore thymeleaf expressions
+      String name = attribute.getName();
+      if (name.startsWith("th:")) {
+        continue;
+      }
```
thomas-serre-sonarsource, line 130: Maybe I should ignore thymeleaf expressions more globally. For instance, in the definition of :
 ``` java 
isUnifiedExpression(value) //line 135
```
But I was not sure of the side effect. Let me know if you have an opinion on this.

Comment (christophe-zurn-sonarsource) [reply sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/scripting/UnifiedExpressionCheck.java:130]: I think it's OK to to ingore thmyleaf expressions globally: the 2 other places where `isUnifiedExpression` is called actually make sense to keep the current logic for thmyleaf expressions in my opinion

**Review by christophe-zurn-sonarsource — APPROVED** [review]
LGTM.
Left a small suggestion for improving the readability of the loop.

File: sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/scripting/UnifiedExpressionCheck.java [hunk 2]
```diff
@@ -122,6 +122,13 @@ public void startDocument(List<Node> nodes) {
...
+
       String value = attribute.getValue();
       if (value != null) {
         value = value.trim();
```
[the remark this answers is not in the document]
thomas-serre-sonarsource: Looks better, thanks.

PR #227, opened 2022-11-28
SONARHTML-161 FP S5256: Table content included with Thymeleaf th:block and th:insert

thomas-serre-sonarsource:


christophe-zurn-sonarsource (changes requested):
I'm not convinced with the current approach.
Maybe we can discuss it a bit offline, I think the logic around attributes and which node should be looked into should be somewhat changed

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableWithoutHeaderCheck.java
```diff
@@ -37,12 +38,23 @@ private static boolean isTable(TagNode node) {
...
 
   private static boolean isHidden(TagNode node) {
     String ariaHidden = node.getAttribute("aria-hidden");
-    return ariaHidden != null && "TRUE".equalsIgnoreCase(ariaHidden);
+    return "TRUE".equalsIgnoreCase(ariaHidden);
+  }
+
+  private static boolean hasThymeleafTemplate(TagNode node) {
+    return node.getAttributes().stream().anyMatch(TableWithoutHeaderCheck::isAttributeThymeleafTemplated) // Thymeleaf in table attributes ?
+      || node.getChildren().stream().map(TagNode::getAttributes).anyMatch(                                // Thymeleaf in children attributes ?
```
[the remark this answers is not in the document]
thomas-serre-sonarsource: Thanks for spotting this. Here is my consideration on this:
- If `th:insert` or `th:include` occur as an attribute of `<table>`, they might insert a TH as a child of `<table>` so compliant code.
- If `th:replace` occurs as an attribute of `<table>`, it is out of scope because it will remove`<table>`. I supposed, maybe wrongly,  we shouldn't encounter such code.
- If `th:block` is a child of `<table>` and has `th:replace,th:insert,th:include`, I assumed a TH could be inserted, so it is compliant. Remark that I'm not sure of the behavior of `th:replace` in `th:block`, but I think it is good enough like this.
- If `th:replace` in any child, it could remove the child tag and replace it with a TH, so it is compliant.

Let me know if you agree with this and if this is sufficiently clear in the code

sonar-html-plugin/src/test/resources/checks/TableWithoutHeaderCheck.html
```diff
@@ -77,3 +77,48 @@
...
+
+<table>
+  <div th:include="/blabla.html"></div>
+  <th:block th:replace="/ContainsTRAndTHAndTD.html" />
+</table>
+
+<table>
+  <th:block th:replace="/ContainsTRAndTHAndTD.html" />
+  <div th:include="/blabla.html"></div>
+</table>
```
christophe-zurn-sonarsource, line 124: There's one case that remains to be covered: the case of potentially nested fragments :
```
<table>
  <tr>
    <div th:replace="/blabl.html"></div>
  </tr>
</table>
```

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableWithoutHeaderCheck.java
```diff
@@ -37,12 +42,32 @@ private static boolean isTable(TagNode node) {
...
+  private static boolean hasThymeleafFragmentInsertionFromTableChildren(List<TagNode> nodes) {
+    boolean hasInsertion = false;
+    for (TagNode node : nodes) {
+      if (node.getNodeName().startsWith("th:block")) {
+        hasInsertion = hasInsertion || node.getAttributes().stream().map(Attribute::getName).anyMatch(THYMELEAF_FRAGMENT_INSERTION_KEYWORDS::contains);
+      } else {
+        hasInsertion = hasInsertion || node.getAttributes().stream().map(Attribute::getName).anyMatch(attributeName -> attributeName.equals("th:replace"));
+      }
+    }
+    return hasInsertion;
```
christophe-zurn-sonarsource, line 67: Note that here we don't necessarily need to go through all the elements and have the `hasInsertion` boolean value be the incisive OR of all values: as soon as one of the condition is true, we could directly return true

christophe-zurn-sonarsource (changes requested):
Nice implementation :)
There's one case remaining the we might want to support: in case the fragments inclusion are not the direct children of the table tag.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableWithoutHeaderCheck.java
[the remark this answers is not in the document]
thomas-serre-sonarsource: After consideration and discussion, it becomes difficult to split the logic between `th:replace` and `th:insert`.It would require to identify each tag name which can include a `<th></th>` in a proper way. 

However, `TableWithoutHeaderCheck` is not that smart for non-template HTML : The code is compliant as soon as it finds a `<th></th>`. I don't think doing something more imaginative for thymeleaf templates is worth it. 

So I simplified the logic : If we insert a thymeleaf fragment using `th:replace` or `th:insert` or `th:include` anywhere inside the hierarchy, I assume the code is compliant.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableWithoutHeaderCheck.java
thomas-serre-sonarsource: Yep, good catch. I've added a helper function, wrapped the conditions in one if, and I will return as soon as possible.

sonar-html-plugin/src/test/resources/checks/TableWithoutHeaderCheck.html
thomas-serre-sonarsource: I've added a test for that.

christophe-zurn-sonarsource (approved):
Nice. LGTM !

PR #228, opened 2022-11-29
SONARHTML-164 Update and remove unneeded dependencies

christophe-zurn-sonarsource:


sonar-html-plugin/src/main/java/org/sonar/plugins/html/node/Node.java
```diff
@@ -39,7 +37,7 @@ protected Node(NodeType nodeType) {
   }
 
   public String getCode() {
-    return code;
+    return code == null ? "" : code;
   }
```
thomas-serre-sonarsource, line 41: _minor_ I'd rather handle the `null` value in the setter `setCode` :
```java
  public void setCode(String code) {
    this.code = code == null ? "" : code;
  }
```
Else, you could have a null value returned by the `toString` function.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/node/Node.java
```diff
@@ -51,7 +49,7 @@ public int getEndLinePosition() {
...
 
   public int getLinesOfCode() {
-    return StringUtils.countMatches(code, "\n");
+    return (int) getCode().chars().filter(charCode -> charCode == '\n').count();
```
thomas-serre-sonarsource, line 52: _minor_
```suggestion
    return (int) getCode().lines().count();
```

sonar-html-plugin/src/main/java/org/sonar/plugins/html/node/TagNode.java
```diff
@@ -102,16 +101,17 @@ public List<TagNode> getChildren() {
...
+    if (localPart.isEmpty()) {
+      return getNodeName();
     } else {
       return localPart;
     }
   }
 
   public String getNodeName() {
-    return nodeName;
+    return nodeName == null ? "" : nodeName;
```
thomas-serre-sonarsource, line 114: _minor_ Same as for `Node`, I'd would rather put this in the setter.

sonar-html-plugin/src/test/java/org/sonar/plugins/html/core/HtmlSensorTest.java
```diff
@@ -72,7 +72,8 @@ public class HtmlSensorTest {
 
   @Before
   public void setUp() {
-    HtmlRulesDefinition rulesDefinition = new HtmlRulesDefinition();
+    final SonarRuntime sonarRuntime = SonarRuntimeImpl.forSonarQube(Version.create(8, 9), SonarQubeSide.SCANNER, SonarEdition.COMMUNITY);
```
thomas-serre-sonarsource, line 75: _curiosity_ It is not the first time I have seen a hardcoded version in tests. Why shouldn't we put this in a global variable or a config file?

thomas-serre-sonarsource (approved):

sonar-html-plugin/pom.xml
```diff
@@ -42,20 +41,7 @@
...
-          <artifactId>sonar-channel</artifactId>
-        </exclusion>
-      </exclusions>
-    </dependency>
-    <!-- jsr305 is provided as its javax.annotation package is shaded inside sonar-analyzer-commons -->
-    <dependency>
-      <groupId>com.google.code.findbugs</groupId>
-      <artifactId>jsr305</artifactId>
-      <version>3.0.2</version>
+      <version>${sonar.plugin.api.version}</version>
```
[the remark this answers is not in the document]
christophe-zurn-sonarsource: We could indeed update to a more recent major version.
The thing is that updating the version of the API will also update the compatibility version with SonarLint.
Given that there was no need to use some new API in the analyzer, I left it as it was.

pom.xml
```diff
@@ -63,11 +63,17 @@
     <license.mailto>sonarqube@googlegroups.com</license.mailto>
     <gitRepositoryName>sonar-html</gitRepositoryName>
 
+    <sslr.version>1.24.0.633</sslr.version>
+    <analyzerCommons.version>2.1.0.1111</analyzerCommons.version>
+    <sonar.plugin.api.version>7.9</sonar.plugin.api.version>
+
+    <sonarlint.api.impl.version>6.3.0.36253</sonarlint.api.impl.version>
+    <sonarqube.api.impl.version>9.7.1.62043</sonarqube.api.impl.version>
+    <sonarqube.sonar.ws.version>9.7.1.62043</sonarqube.sonar.ws.version>
```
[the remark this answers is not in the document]
christophe-zurn-sonarsource: These are the version of the implementation of the API mentioned above, these are only used inside tests.
No specific process, we can update them whenever needed.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/node/Node.java
christophe-zurn-sonarsource: Much better. Thanks !

sonar-html-plugin/src/main/java/org/sonar/plugins/html/node/Node.java
christophe-zurn-sonarsource: Good point for the `toString` method. I'll simply call `getCode` inside it instead, if it makes sense with you

sonar-html-plugin/src/test/java/org/sonar/plugins/html/core/HtmlSensorTest.java
christophe-zurn-sonarsource: No reason :D
I can indeed put it in a constant somewhere. I just updated the test one after the other and did not think of it :+1:

sonar-html-plugin/src/main/java/org/sonar/plugins/html/node/TagNode.java
christophe-zurn-sonarsource: (see comment about `setCode`, it s a similar reasoning)

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/scripting/UnifiedExpressionCheck.java
```diff
@@ -68,7 +68,7 @@ public class UnifiedExpressionCheck extends AbstractPageCheck {
     defaultValue = DEFAULT_FUNCTIONS)
   public String functions = DEFAULT_FUNCTIONS;
 
-  private String[] functionsArray;
+  private Set<String> functionsSet = Set.of();
```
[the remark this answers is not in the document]
christophe-zurn-sonarsource: I would really avoid using interfaces such as `AbstractImmutableSet` unless really necessary.
I'd prefer to initialize the variable, as what was there before:
`ArrayUtils.contains(functionsArray, localName)` was resilient against nullable values, so I wanted to maintain the same defensive behavior.
If you don't like the immutability, how about simply initiliazing an empty hash set instead ? `Set<String> functionsSet = new HashSet<>();`? WDYT?

sonar-html-plugin/pom.xml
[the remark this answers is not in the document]
thomas-serre-sonarsource: sounds good! thanks

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/scripting/UnifiedExpressionCheck.java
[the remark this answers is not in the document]
thomas-serre-sonarsource: I upvote your last proposition : 
```java 
Set<String> functionsSet = new HashSet<>(); 
```

Thanks :)

sonar-html-plugin/src/test/java/org/sonar/plugins/html/core/HtmlSensorTest.java
thomas-serre-sonarsource: Alright !

pom.xml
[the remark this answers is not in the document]
thomas-serre-sonarsource: Ok, thanks for the answer !

sonar-html-plugin/src/main/java/org/sonar/plugins/html/node/Node.java
thomas-serre-sonarsource: Oh, my bad, I didn't think about the possibility of the variable not being initialized 🙄. Thanks for the explanation.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/node/Node.java
christophe-zurn-sonarsource: As discussed offline, this change would actually introduce a change in the underlying logic. While this would fix a wrong behavior, it's out of the scope of the current ticket.
I've created https://sonarsource.atlassian.net/browse/SONARHTML-165 to tackle this separately in our current sprint.

PR #229, opened 2022-11-30
SONARHTML-143 Web:UnifiedExpressionCheck should support maps

christophe-zurn-sonarsource:


thomas-serre-sonarsource (approved):
Looks good! simple and efficient

PR #230, opened 2022-11-30
SONARHTML-158 FP and FN for multiple rules: Support Angular attribute binding

thomas-serre-sonarsource:


christophe-zurn-sonarsource (approved):
LGTM ! Nice

PR #231, opened 2022-12-02
SONARHTML-165 Fix 'Node#getLinesOfCode' method and adapt Web:LongJavaScriptCheck

thomas-serre-sonarsource:


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/scripting/LongJavaScriptCheck.java
```diff
@@ -28,39 +28,46 @@ public class LongJavaScriptCheck extends AbstractPageCheck {
...
     defaultValue = "" + DEFAULT_MAX_LINES)
   public int maxLines = DEFAULT_MAX_LINES;
 
-  private int linesOfCode;
-  private TagNode scriptNode;
+  @Override
+  public void startElement(TagNode node) {
+    if ("script".equalsIgnoreCase(node.getNodeName())) {
+      scriptNode = node;
+      startingNodeLinePosition = node.getStartLinePosition();
```
christophe-zurn-sonarsource, line 47: Given that we already store the `node` as a field, we don't have to store the start line too, as we can simply use `scriptNode.getStartLinePosition()` in the `endElement` method when needed.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/scripting/LongJavaScriptCheck.java
```diff
@@ -28,39 +28,46 @@ public class LongJavaScriptCheck extends AbstractPageCheck {
...
-    if (scriptNode != null) {
-      linesOfCode += textNode.getLinesOfCode();
-
-      if (linesOfCode > maxLines) {
-        createViolation(scriptNode.getStartLinePosition(),
-          "The length of this JS script (" + linesOfCode + ") exceeds the maximum set to " + maxLines + ".",
-          Double.valueOf(linesOfCode) - Double.valueOf(maxLines));
-        scriptNode = null;
-      }
+    if(scriptNode != null){
```
christophe-zurn-sonarsource, line 54: _minor_ this line as a weird styling
```suggestion
    if (scriptNode != null) {
```

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/scripting/LongJavaScriptCheck.java
```diff
@@ -28,39 +28,46 @@ public class LongJavaScriptCheck extends AbstractPageCheck {
...
-  }
+    if ("script".equalsIgnoreCase(element.getNodeName()) && scriptNode != null) {
 
-  @Override
-  public void startElement(TagNode element) {
-    if ("script".equalsIgnoreCase(element.getNodeName())) {
-      scriptNode = element;
-      linesOfCode = 0;
+      // We dot not consider 1st line and last line, where <script> tags are
+      int linesOfCode = (int) (text.toString().trim().lines().count());
```
christophe-zurn-sonarsource, line 64: I think we should actually remove this comment as it is misleading: we decided to trim the string in the end, instead of not considering 1st line and the last line as said in the comment.

christophe-zurn-sonarsource, line 64: _minor_ there are unneeded parenthesis here:
```suggestion
      // We dot not consider 1st line and last line, where <script> tags are
      int linesOfCode = (int) text.toString().trim().lines().count();
```

christophe-zurn-sonarsource (changes requested):

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/scripting/LongJavaScriptCheck.java
thomas-serre-sonarsource: Great catch! Automatic reformating doesn't want to catch it weirdly.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/scripting/LongJavaScriptCheck.java
thomas-serre-sonarsource: yep, agreed

sonar-html-plugin/src/main/java/org/sonar/plugins/html/analyzers/PageCountLines.java
```diff
@@ -71,71 +71,35 @@ private void count(List<Node> nodeList) {
...
         break;
       default:
         break;
     }
   }
 
-  private void handleTokenComment(Node node, @Nullable Node previousNode, int linesOfCodeCurrentNode) {
-    if (previousNode == null) {
-      // this is a header comment
-      headerCommentLines += linesOfCodeCurrentNode;
```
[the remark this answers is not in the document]
thomas-serre-sonarsource: Indeed.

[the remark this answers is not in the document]
thomas-serre-sonarsource: I did altogether remove the LOGGER from this class. Do you agree?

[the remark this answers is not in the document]
christophe-zurn-sonarsource: Yes, given that it was only for a single debug log, it's fine by me :D :+1:

christophe-zurn-sonarsource (approved):
LGTM !
I restarted the CI given it was broken on Friday. We can merge once it passes :)

PR #232, opened 2022-12-02
SONARHTML-135 SonarHTML should not depend on sonar-channel

christophe-zurn-sonarsource:


sonar-html-plugin/src/main/java/org/sonar/plugins/html/lex/AbstractTokenizer.java
```diff
@@ -99,6 +98,14 @@ public boolean consume(CodeReader codeReader, T nodeList) {
     }
   }
 
+  protected static String popTo(CodeReader codeReader, EndMatcher endMatcher, StringBuilder stringBuilder) {
+    do {
+      stringBuilder.append((char) codeReader.pop());
+    } while (!endMatcher.match(codeReader.peek()) && codeReader.peek() != -1);
```
thomas-serre-sonarsource, line 104: Shouldn't you start with a `while` rather than a `do while`:
 ```suggestion
    while (!endMatcher.match(codeReader.peek()) && codeReader.peek() != -1) {
      stringBuilder.append((char) codeReader.pop());
    }
```
In the case that `codeReader.peek()` returns -1 at the first iteration ?

thomas-serre-sonarsource (changes requested):

sonar-html-plugin/src/main/java/org/sonar/plugins/html/lex/AbstractTokenizer.java
christophe-zurn-sonarsource: Indeed we can do this. But we have to do it a bit differently, as we still need to pop the first character before starting to test for the `endMatcher`

thomas-serre-sonarsource: Oh ok got it keep it like this then :).

thomas-serre-sonarsource (approved):

sonar-html-plugin/src/main/java/org/sonar/plugins/html/lex/AbstractTokenizer.java
thomas-serre-sonarsource: All good! Thanks for the modification

thomas-serre-sonarsource:
Looks good to me!

PR #241, opened 2023-01-31
fix CI java version

ilia-kebets-sonarsource:


ilia-kebets-sonarsource:
QA is now failing because:

```
> java.lang.ClassCastException: class java.io.ObjectStreamClass$Caches$1 cannot be cast to class java.util.Map (java.io.ObjectStreamClass$Caches$1 and java.util.Map are in module java.base of loader 'bootstrap')
```

saberduck (approved):

PR #257, opened 2023-11-21
SONARHTML-183 Rule S5247: Disabling auto-escaping in template engines is security-sensitive

guillaume-dequenne-sonarsource:


maksim-grebeniuk-sonarsource (approved):
LGTM, just one thing I think it makes sense to have - is test case when there is more than one `autoescape false` in a single text node

sonar-html-plugin/src/main/resources/org/sonar/l10n/web/rules/Web/Sonar_way_profile.json
```diff
@@ -25,6 +25,7 @@
     "S5260",
     "S5264",
     "S5725",
+    "S5247",
```
saberduck, line 28: let's keep this sorted

saberduck (changes requested):
see the comment, rule keys in profile json file should be sorted

saberduck (approved):

PR #261, opened 2024-01-25
SONARHTML-189 Require SonarHTML download on JavaScript and PHP languages

yassin-kammoun-sonarsource:


ilia-kebets-sonarsource (commented):
I'm a bit confused as to why you are deleting these tests. Should updating the `requiredForLanguages` property fix the lack of analysis that was failing them in the first place?

ilia-kebets-sonarsource:
So, with the new setup, for files of the Sonar `php` and `js` languages to be analyzed, their responsible plugin (SonarJS and SonarPHP) are required. Without them, they will not be analyzed by the other analyzers that also lint them, such as here SonarHTML.
We have decided to remove a few tests because this behaviour is verified elsewhere.

ilia-kebets-sonarsource (approved):
LGTM

PR #264, opened 2024-02-13
SONARHTML-196 Create rule S6846: DOM elements should not use the `accesskey` property

ilia-kebets-sonarsource:
RSPEC PR: https://github.com/SonarSource/rspec/pull/3646


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/attributes/NoAccessKeyCheck.java
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.sonar.plugins.html.checks.attributes;
+
+import org.sonar.check.Rule;
+
+@Rule(key = "S6846")
+public class NoAccessKeyCheck extends IllegalAttributeCheck {
```
yassin-kammoun-sonarsource, line 23: I am not sure it's a good idea to extend `IllegalAttributeCheck`. It defines a rule property that `NoAccessKeyCheck` would likely also inherit.

Furthermore, I am not a big fan of having rules extending others. From experience with CDK sprints, abstracting over rule patterns becomes painful in the long run, especially when we need to go back and make some changes that could impact more than one rule unexpectedly.

sonar-html-plugin/src/main/resources/org/sonar/l10n/web/rules/Web/S6846.json
```diff
@@ -0,0 +1,24 @@
...
+    "constantCost": "5min"
+  },
+  "tags": [
+    "accessibility"
+  ],
+  "defaultSeverity": "Minor",
+  "ruleSpecification": "RSPEC-6846",
+  "sqKey": "S6846",
+  "scope": "All",
+  "quickfix": "targeted",
```
yassin-kammoun-sonarsource: As mentioned in the RSPEC PR, the rule doesn't provide a quickfix. Let's not bother for now (unless you want to give it a try).

yassin-kammoun-sonarsource (changes requested):

yassin-kammoun-sonarsource:
> Fixes https://sonarsource.atlassian.net/browse/SONARHTML-196 RSPEC PR: [SonarSource/rspec#3646](https://github.com/SonarSource/rspec/pull/3646)

For your information, because we are using Jira, there is no need to add `Fixes ...`. Instead, you just need to prefix your PR title with SONARHTML-196. You could even do the same with commit messages. It happens to be a convention for all repositories integrated with Jira.

For example, see sonar-python:

- https://github.com/SonarSource/sonar-python/commits/master/

PR #266, opened 2024-02-14
Document command to update ruling results

ilia-kebets-sonarsource:


README.md
```diff
@@ -53,6 +53,10 @@ The "Ruling Test" are an integration test suite which launches the analysis of a
     cd its/ruling
     mvn verify
 
+To update the ruling results run:
+
+    rsync -avh its/ruling/target/actual/ its/ruling/src/test/resources/expected/ --delete
```
yassin-kammoun-sonarsource, line 58: Do you think we could add a Maven command in the ruling `pom.xml` like `mvn update` that would execute this bash command?

PR #267, opened 2024-02-14
SONARHTML-207 Create rule S6827: Anchors should contain accessible content

yassin-kammoun-sonarsource:
Inspired from [anchor-has-content](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/rules/anchor-has-content.js)


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/AnchorsHaveContentCheck.java
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  private static boolean hasContent(TagNode element) {
+    var children = element.getChildren();
+    for (TagNode child : children) {
+      if (!isHidden(child)) {
+        return true;
+      }
+    }
+    if (element.hasProperty("title") || element.hasProperty("aria-label")) {
```
ilia-kebets-sonarsource, line 103: why not refactor this [as Sonar suggests](https://next.sonarqube.com/sonarqube/project/issues?pullRequest=267&issueStatuses=ACCEPTED&open=878bde4e-41cb-4721-b81c-a00eb693b65a&id=org.sonarsource.html%3Ahtml)?

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/AnchorsHaveContentCheck.java
```diff
@@ -0,0 +1,114 @@
...
+      return true;
+    }
+    return false;
+  }
+
+  private static boolean isHidden(TagNode element) {
+    return ("input".equalsIgnoreCase(element.getNodeName())
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: It's come from here:

https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/0d5321a5457c5f0da0ca216053cc5b4f571b53ae/src/rules/anchor-has-content.js#L42C13-L42C31

which leads to:

https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/util/hasAccessibleChild.js#L7

which eventually leads to:

https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/util/isHiddenFromScreenReader.js#L10

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/AnchorsHaveContentCheck.java
yassin-kammoun-sonarsource: Sure.

ilia-kebets-sonarsource:
Why not capture the following cases:
```html
<a />
<a><Bar aria-hidden /></a> <!--this one will need to be adapted to something that exists in HTML>
```

I tested all from https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/__tests__/src/rules/anchor-has-content-test.js

ilia-kebets-sonarsource (approved):
[Last comment](https://github.com/SonarSource/sonar-html/pull/267#issuecomment-1944143848), otherwise LGTM!

yassin-kammoun-sonarsource:
> Why not capture the following cases:
> 
> ```
> <a />
> <a><Bar aria-hidden /></a> <!--this one will need to be adapted to something that exists in HTML>
> ```
> 
> I tested all from https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/__tests__/src/rules/anchor-has-content-test.js

I am not sure about `<a><Bar aria-hidden /></a>` as the [undefined value is unpredictable](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden#values).

H-Lo:
screenshot

https://community.sonarsource.com/t/anchors-must-have-content-and-the-content-must-be-accessible-by-a-screen-reader-web-s6827/111545

PR #269, opened 2024-02-14
SONARHTML-191 Create rule S6850: Heading elements should have accessible content

ericmorand-sonarsource:


sonar-html-plugin/src/main/resources/org/sonar/l10n/web/rules/Web/S6850.json
```diff
@@ -0,0 +1,25 @@
...
+  "title": "Heading elements should have accessible content",
+  "type": "CODE_SMELL",
+  "status": "ready",
+  "remediation": {
+    "func": "Constant\/Issue",
+    "constantCost": "5min"
+  },
+  "tags": [
+    "accessibility",
+    "react"
```
ilia-kebets-sonarsource, line 11: run rule-api to update this

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/HeadingHasAccessibleContentCheck.java
```diff
@@ -0,0 +1,137 @@
...
+    // vueJS attributes that maps to content are considered as content
+    vueJsContentLikeAttributes.forEach(attributeName -> {
+      String nodeAttribute = node.getAttribute(attributeName);
+
+      if (nodeAttribute != null && !nodeAttribute.isBlank() && bufferStack.getLevel() > 0) {
+        bufferStack.write(nodeAttribute);
+      }
+    });
+
+    super.startElement(node);
```
ilia-kebets-sonarsource, line 77: do we need to call super?

ilia-kebets-sonarsource:
I added a prefix to your PR because https://github.com/SonarSource/sonar-html/pull/264#issuecomment-1943521319

sonar-html-plugin/src/main/resources/org/sonar/l10n/web/rules/Web/S6850.json
ericmorand-sonarsource: Done

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/HeadingHasAccessibleContentCheck.java
ericmorand-sonarsource: I don't know. Is there a convention with Java - or at SonarSource - that allows us to know if we must call a method parent's implementation?

yassin-kammoun-sonarsource: It's not really about convention but rather about intention and needs:

- In a constructor, you _must_ call the parent constructor to first initialize its state before proceeding with the child's. 
- In a method, you _could_ call the parent's implementation of the method if you need to trigger a side-effect you need.

I don't think it makes sense to make these calls, especially when the parent does nothing:

https://github.com/SonarSource/sonar-html/blob/4de7dfe5ceb1e7964f7a31e2ff8270a5b0bae8c5/sonar-html-plugin/src/main/java/org/sonar/plugins/html/visitor/DefaultNodeVisitor.java#L68-L70

ericmorand-sonarsource: Yes, I will remove them because, here, we _know_ the parent's implementations do nothing.

> In a method, you could call the parent's implementation of the method if you need to trigger a side-effect you need.

Can I assume that, if the parent's implementation has side-effect (or _must_ be called), it would be documented in the parent class?

sonar-html-plugin/src/test/resources/checks/HeadingHasAccessibleContentCheck.html
```diff
@@ -0,0 +1,92 @@
+<h1>Compliant: h1 with a content</h1>
+<h2>Compliant: h2 with a content</h2>
+<h3>Compliant: h3 with a content</h3>
+<h4>Compliant: h4 with a content</h4>
+<h5>Compliant: h5 with a content</h5>
+<h6>Compliant: h6 with a content</h6>
+
+<h1 aria-hidden>Non-compliant: h1 with an aria-hidden attribute</h1>
```
ilia-kebets-sonarsource, line 8: can you add a comment that those are noncompliant?

sonar-html-plugin/src/test/resources/checks/HeadingHasAccessibleContentCheck.html
```diff
@@ -0,0 +1,92 @@
...
+<h2 aria-hidden>Non-compliant: h2 with an aria-hidden attribute</h2>
```
ilia-kebets-sonarsource, line 9: `Noncompliant` is not hyphenated

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/HeadingHasAccessibleContentCheck.java
```diff
@@ -0,0 +1,125 @@
...
+
+  @Override
+  public void expression(ExpressionNode expressionNode) {
+    if (bufferStack.getLevel() > 0) {
+      bufferStack.write(expressionNode.toString());
+    }
+  }
+
+  @Override
+  public void directive(DirectiveNode directiveNode) {
```
ilia-kebets-sonarsource, line 110: Is it not necessary to check that the directive starts with `?php`?

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/HeadingHasAccessibleContentCheck.java
yassin-kammoun-sonarsource: Mmm, well if it's the case and we have to, sure, but I would first find the author of the parent's class, send them to Mordor, throw them to Mount Doom, and then, if they are still alive give a 1-hour lecture about inheritance, polymorphism, and encapsulation.

sonar-html-plugin/src/test/java/org/sonar/plugins/html/checks/accessibility/HeadingHasAccessibleContentCheckTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+import java.io.File;
+
+class HeadingHasAccessibleContentCheckTest {
+  @RegisterExtension
+  public CheckMessagesVerifierRule checkMessagesVerifier = new CheckMessagesVerifierRule();
+
+  @Test
+  void html() throws Exception {
+    HtmlSourceCode sourceCode = TestHelper.scan(new File("src/test/resources/checks/HeadingHasAccessibleContentCheck.html"), new HeadingHasAccessibleContentCheck());
```
ilia-kebets-sonarsource, line 34: Could you move the fixture files into `src/test/resources/checks/HeadingHasAccessibleContentCheck/`:
- `src/test/resources/checks/HeadingHasAccessibleContentCheck/file.html`
- `src/test/resources/checks/HeadingHasAccessibleContentCheck/file.jsp`
- `src/test/resources/checks/HeadingHasAccessibleContentCheck/file.php`
- `src/test/resources/checks/HeadingHasAccessibleContentCheck/file.vue`

ilia-kebets-sonarsource (approved):
See my remaining comments, otherwise LGTM!

PR #270, opened 2024-02-15
SONARHTML-194 Create rule S6851: Images should have a non-redundant alternate description

yassin-kammoun-sonarsource:
Inspired from [img-redundant-alt](https://github.com/bradbirdsallCHANGED/eslint-plugin-jsx-a11y/blob/main/src/rules/img-redundant-alt.js)


ilia-kebets-sonarsource (approved):
LGTM!

PR #272, opened 2024-02-16
Fix S6846: location on tag name was 1 column too far

ilia-kebets-sonarsource:
Screenshot 2024-02-16 at 09 28 13


yassin-kammoun-sonarsource (approved):
Looks good to me, but to be honest, I wonder whether we should avoid doing this kind of error-prone arithmetic and keep raising on nodes and lines just like the other rules do for now.

For me, it would make more sense to plan a different sprint and invest in improving the analyzer API for better issue reporting (and maybe quickfixes). Then we could come back to each implemented rule and improve reported issue locations.

What do you think?

PR #273, opened 2024-02-19
SONARHTML-202 Create rule S6842: Non-interactive DOM elements should not have interactive ARIA roles

ilia-kebets-sonarsource:
We've expanded known tags with `html` and [deprecated tags](https://developer.mozilla.org/en-US/docs/Web/HTML/Element#obsolete_and_deprecated_elements).

To generate the list of interactive roles, I have simply run [this code](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/0d5321a5457c5f0da0ca216053cc5b4f571b53ae/src/util/isInteractiveRole.js#L9-L12).

To generate the list of non-interactive elements, I have applied [this filter](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/0d5321a5457c5f0da0ca216053cc5b4f571b53ae/src/util/isNonInteractiveElement.js#L145) to the list of known tags.


sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/HtmlConstants.java
```diff
@@ -141,13 +161,25 @@ public class HtmlConstants {
...
+    "wbr",
+    "xmp" // deprecated
   );
 
+  // computed from https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/util/isNonInteractiveElement.js
+  public static final Set<String> NON_INTERACTIVE_ELEMENTS = Set.of("abbr", "address", "article", "aside", "blockquote", "br", "caption", "code", "dd", "del", "details", "dfn",
+    "dialog", "dir", "dl", "dt", "em", "fieldset", "figcaption", "figure", "footer", "form", "h1", "h2", "h3", "h4", "h5", "h6", "hr", "html", "iframe", "img", "ins", "label",
+    "legend", "li", "main", "mark", "marquee", "menu", "meter", "nav", "ol", "optgroup", "output", "p", "pre", "progress", "ruby", "strong", "sub", "sup", "table", "tbody",
+    "tfoot", "thead", "time", "ul");
+  // computed as https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/util/isInteractiveRole.js
```
yassin-kammoun-sonarsource, line 178:
```suggestion

  // computed as https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/util/isInteractiveRole.js
```

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/HtmlConstants.java
```diff
@@ -141,13 +161,25 @@ public class HtmlConstants {
...
     "ul",
     "var",
     "video",
-    "wbr"
+    "wbr",
+    "xmp" // deprecated
   );
 
+  // computed from https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/util/isNonInteractiveElement.js
+  public static final Set<String> NON_INTERACTIVE_ELEMENTS = Set.of("abbr", "address", "article", "aside", "blockquote", "br", "caption", "code", "dd", "del", "details", "dfn",
```
yassin-kammoun-sonarsource, line 174: Could you add another set for interactive elements?

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/HtmlConstants.java
```diff
@@ -141,13 +161,25 @@ public class HtmlConstants {
...
+    "dialog", "dir", "dl", "dt", "em", "fieldset", "figcaption", "figure", "footer", "form", "h1", "h2", "h3", "h4", "h5", "h6", "hr", "html", "iframe", "img", "ins", "label",
+    "legend", "li", "main", "mark", "marquee", "menu", "meter", "nav", "ol", "optgroup", "output", "p", "pre", "progress", "ruby", "strong", "sub", "sup", "table", "tbody",
+    "tfoot", "thead", "time", "ul");
+  // computed as https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/util/isInteractiveRole.js
+  public static final Set<String> INTERACTIVE_ROLES = Set.of("button", "checkbox", "columnheader", "combobox", "grid", "gridcell", "link", "listbox", "menu", "menubar", "menuitem",
```
yassin-kammoun-sonarsource, line 179: Could you add another set for noninteractive roles?

yassin-kammoun-sonarsource (approved):
I suggest we already add sets for interactive elements and nonintereactive roles since you introduce their complement in this PR.

Looks good to me otherwise!

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/HtmlConstants.java
ilia-kebets-sonarsource: Isn't it simply?
```java
! NON_INTERACTIVE_ELEMENTS.contains(myTag)
```

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/HtmlConstants.java
ilia-kebets-sonarsource: similar to https://github.com/SonarSource/sonar-html/pull/273/files#r1495810809

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/HtmlConstants.java
yibnl: Indeed, never mind my comment.

PR #274, opened 2024-02-19
SONARHTML-215 Create rule S6793: ARIA properties in DOM elements should have valid values

yassin-kammoun-sonarsource:


sonar-html-plugin/src/test/resources/checks/AriaProptypesCheck.html
```diff
@@ -0,0 +1,53 @@
+<!-- Non-aria attribute -->
+<div aria-foo="true" />
+<div foo="true" />
+
+<!-- Boolean -->
+<div aria-hidden="true" />
+<div aria-hidden="false" />
+<div aria-hidden="undefined" />
```
ilia-kebets-sonarsource, line 8: should we support `undefined` in HTML? I mean it makes sense in JSX, but here?

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/Aria.java
```diff
@@ -0,0 +1,147 @@
...
+    ARIA_PROPERTIES.put("aria-valuemin", new AriaProperty("aria-valuemin", AriaPropertyType.NUMBER));
+    ARIA_PROPERTIES.put("aria-valuenow", new AriaProperty("aria-valuenow", AriaPropertyType.NUMBER));
+    ARIA_PROPERTIES.put("aria-valuetext", new AriaProperty("aria-valuetext", AriaPropertyType.STRING));
+  }
+
+  static class AriaProperty {
+    private final String name;
+    private final AriaPropertyType type;
+    private final Optional<Boolean> allowUndefined;
+    private final List<String> values;
```
ilia-kebets-sonarsource, line 88: why not use a Set?

sonar-html-plugin/src/test/resources/checks/AriaProptypesCheck.html
yassin-kammoun-sonarsource: Indeed, it doesn't make any sense here.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/Aria.java
yassin-kammoun-sonarsource: Sure, but I won't be able to assert some issue messages based on this set as the value order won't be deterministic on collection.

ilia-kebets-sonarsource: You did the change, meaning it's all good?

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/AriaProptypesCheck.java
```diff
@@ -0,0 +1,108 @@
...
+
+      var value = attribute.getValue();
+      var property = ARIA_PROPERTIES.get(normalizedName);
+      if (!isValid(property, value)) {
+        createViolation(element.getStartLinePosition(), message(name, property.getType(), property.getValues()));
+      }
+    }
+  }
+
+  private static boolean isValid(AriaProperty property, String value) {
```
ilia-kebets-sonarsource, line 54: These might be candidates to be moved to the `Aria.java` helper class.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/Aria.java
yassin-kammoun-sonarsource: Yes

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/AriaProptypesCheck.java
yassin-kammoun-sonarsource: I don't think we need to. This is really specific to the rule, and the JavaScript implementation doesn't expose it as a helper function anywhere outside the rule.

PR #276, opened 2024-02-20
SONARHTML-203 Create rule S6843: Interactive DOM elements should not have non-interactive ARIA roles

ilia-kebets-sonarsource:


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoInteractiveElementToNoninteractiveRoleCheck.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @Override
+  public void startElement(TagNode node) {
+    var role = node.getPropertyValue("role");
+    var tagName = node.getNodeName().toLowerCase(Locale.ROOT);
+    if (
+      role != null &&
+      KNOWN_HTML_TAGS.contains(tagName) &&
+      INTERACTIVE_ELEMENTS.contains(tagName) &&
+      (NON_INTERACTIVE_ROLES.contains(role) || PRESENTATION_ROLES.contains(role))
```
ilia-kebets-sonarsource, line 43: `PRESENTATION_ROLES` is a subset of `NON_INTERACTIVE_ROLES`. We could simplify this, but we would be diverging from the eslint-plugin logic. WDYT?

yassin-kammoun-sonarsource: I think it's okay to have distinct sets in case one happens to change after getting user feedback.

sonar-html-plugin/src/test/java/org/sonar/plugins/html/checks/accessibility/NoInteractiveElementToNoninteractiveRoleCheckTest.java
```diff
@@ -0,0 +1,117 @@
...
+  @Test
+  void invalid() throws Exception {
+    HtmlSourceCode sourceCode = TestHelper.scan(
+      new File("src/test/resources/checks/NoInteractiveElementToNoninteractiveRoleCheck/invalid.html"),
+      new NoInteractiveElementToNoninteractiveRoleCheck());
+
+    checkMessagesVerifier.verify(sourceCode.getIssues())
+      .next().atLine(2).withMessage("Interactive elements should not be assigned non-interactive roles.")
+      .next().atLine(3)
+      //.next().atLine(4)
```
yassin-kammoun-sonarsource, line 50: Why was this left commented?

sonar-html-plugin/src/test/java/org/sonar/plugins/html/checks/accessibility/NoInteractiveElementToNoninteractiveRoleCheckTest.java
```diff
@@ -0,0 +1,117 @@
...
+      .next().atLine(20)
+      .next().atLine(21)
+      .next().atLine(22)
+      .next().atLine(23)
+      .next().atLine(24)
+      .next().atLine(25)
+      .next().atLine(26)
+      .next().atLine(27)
+      .next().atLine(28)
+      //.next().atLine(29)
```
yassin-kammoun-sonarsource, line 75: There are more.

yassin-kammoun-sonarsource (approved):
Looks good to me, but there are test assertions that are commented.
We need to address this before merging; either leave a comment explaining why or just drop them if they are irrelevant.

sonar-html-plugin/src/test/java/org/sonar/plugins/html/checks/accessibility/NoInteractiveElementToNoninteractiveRoleCheckTest.java
ilia-kebets-sonarsource: these are the lines that have comments in the fixtures. It seemed more readable that way, instead of just removing them. WDYT?

yassin-kammoun-sonarsource: YAGNI.

PR #277, opened 2024-02-21
SONARHTML-200 Create rule S6847: Non-interactive elements shouldn't have event handlers

yassin-kammoun-sonarsource:


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoNonInteractiveElementsWithHandlersCheck.java
```diff
@@ -0,0 +1,109 @@
...
+    var role = element.getAttribute("role");
+    return role != null && INTERACTIVE_ROLES.stream().anyMatch(role::equalsIgnoreCase);
+  }
+
+  private static boolean isNonInteractiveRole(TagNode element) {
+    var role = element.getAttribute("role");
+    return role != null && NON_INTERACTIVE_ROLES.stream().anyMatch(role::equalsIgnoreCase);
+  }
+
+  private static boolean isPresentationRole(TagNode element) {
```
ilia-kebets-sonarsource, line 100: you could reuse the set I've introduced in S6843

ilia-kebets-sonarsource (approved):
I left a single comment to reuse some code I did in parallel.
Watch out when merging lists from master, I added <a> in interactive elements.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoNonInteractiveElementsWithHandlersCheck.java
yassin-kammoun-sonarsource: 👍

PR #278, opened 2024-02-21
Create rule S6840: DOM elements should use the "autocomplete" attribute correctly

ericmorand-sonarsource:


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck.java
```diff
@@ -0,0 +1,323 @@
...
+import org.sonar.plugins.html.api.accessibility.ControlGroup;
+import org.sonar.plugins.html.checks.AbstractPageCheck;
+import org.sonar.plugins.html.node.TagNode;
+
+import java.util.List;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+
+@Rule(key = "S6840")
+public class DomElementsShouldUseAutocompleteAttributeCorrectlyCheck extends AbstractPageCheck {
```
yassin-kammoun-sonarsource, line 31: Let's rename the check to something that doesn't look like a sentence or sound like the name of a test case :)
I suggest naming it `ValidAutocompleteCheck` (or at least something shorter).

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck.java
```diff
@@ -0,0 +1,323 @@
...
+    ) {
+      this.value = value;
+      this.tagNode = tagNode;
+    }
+
+    public boolean satisfies(Token token) {
+      /*
+       * Directives are considered as wildcards - i.e. they always match the passed token
+       */
+      return (isADirective(this.value) || this.value.equals(token.value)) && token.controlGroupPredicate.test(this.tagNode);
```
yassin-kammoun-sonarsource, line 105:
```suggestion
      return (isADirective(this.value) || this.value.equalsIgnoreCase(token.value)) && token.controlGroupPredicate.test(this.tagNode);
```

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck.java
```diff
@@ -0,0 +1,323 @@
...
+
+        isValid = isValid && constraint.test(candidate);
+      }
+
+      return isValid;
+    }
+
+    int size() {
+      return constraints.size();
+    }
```
yassin-kammoun-sonarsource, line 71: We can inline this method since it's only used once and especially in `isValid`.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck.java
```diff
@@ -0,0 +1,323 @@
...
+      /*
+       * Directives are considered as wildcards - i.e. they always match the passed token
+       */
+      return (isADirective(this.value) || this.value.equals(token.value)) && token.controlGroupPredicate.test(this.tagNode);
+    }
+  }
+
+  static List<Validator> validators = List.of(
+    new Validator(List.of(
+      value -> value.value.isBlank()
```
yassin-kammoun-sonarsource, line 111:
```suggestion
      candidate -> candidate.value.isBlank()
```

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck.java
```diff
@@ -0,0 +1,323 @@
...
+      new Token("webauthn", node -> node.getNodeName().equals(HtmlConstants.NAME_INPUT) || node.getNodeName().equals(HtmlConstants.NAME_TEXTAREA))
+    ).anyMatch(candidate::satisfies);
+  }
+
+  @Override
+  public void startElement(TagNode node) {
+    boolean isValid = validators.stream().anyMatch(validator -> validator.isValid(node));
+
+    if (!isValid) {
+      createViolation(node, "DOM elements should use the \"autocomplete\" attribute correctly");
```
yassin-kammoun-sonarsource, line 320:
```suggestion
      createViolation(node, "DOM elements should use the \"autocomplete\" attribute correctly.");
```

sonar-html-plugin/src/test/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheckTest.java
```diff
@@ -0,0 +1,118 @@
...
+    "new-password",
+    "street-address",
+    "url",
+    "username"
+  })
+  void html(String file) {
+    HtmlSourceCode sourceCode = TestHelper.scan(new File(String.format("src/test/resources/checks/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck/%s.html", file)), new DomElementsShouldUseAutocompleteAttributeCorrectlyCheck());
+
+    checkMessagesVerifier.verify(sourceCode.getIssues())
+      .next().atLine(1).withMessage("DOM elements should use the \"autocomplete\" attribute correctly")
```
yassin-kammoun-sonarsource, line 48:
```suggestion
      .next().atLine(1).withMessage("DOM elements should use the \"autocomplete\" attribute correctly.")
```

sonar-html-plugin/src/test/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheckTest.java
```diff
@@ -0,0 +1,118 @@
...
+      .next().atLine(3);
+  }
+
+  @Test
+  void bday() {
+    DomElementsShouldUseAutocompleteAttributeCorrectlyCheck check = new DomElementsShouldUseAutocompleteAttributeCorrectlyCheck();
+    HtmlSourceCode htmlSourceCode = TestHelper.scan(new File("src/test/resources/checks/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck/bday/file.html"), check);
+
+    checkMessagesVerifier.verify(htmlSourceCode.getIssues())
+      .next().atLine(1).withMessage("DOM elements should use the \"autocomplete\" attribute correctly")
```
yassin-kammoun-sonarsource, line 59:
```suggestion
      .next().atLine(1).withMessage("DOM elements should use the \"autocomplete\" attribute correctly.")
```

sonar-html-plugin/src/test/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheckTest.java
```diff
@@ -0,0 +1,118 @@
...
+    checkMessagesVerifier.verify(jspSourceCode.getIssues());
+
+    HtmlSourceCode phtmlSourceCode = TestHelper.scan(new File("src/test/resources/checks/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck/bday/file.phtml"), check);
+
+    checkMessagesVerifier.verify(phtmlSourceCode.getIssues());
+
+    HtmlSourceCode vueSourceCode = TestHelper.scan(new File("src/test/resources/checks/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck/bday/file.vue"), check);
+
+    checkMessagesVerifier.verify(vueSourceCode.getIssues())
+      .next().atLine(2).withMessage("DOM elements should use the \"autocomplete\" attribute correctly")
```
yassin-kammoun-sonarsource, line 74:
```suggestion
      .next().atLine(2).withMessage("DOM elements should use the \"autocomplete\" attribute correctly.")
```

sonar-html-plugin/src/test/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheckTest.java
```diff
@@ -0,0 +1,118 @@
...
+      .next().atLine(3)
+      .next().atLine(4);
+  }
+
+  @Test
+  void on() {
+    HtmlSourceCode sourceCode = TestHelper.scan(new File("src/test/resources/checks/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck/on.html"), new DomElementsShouldUseAutocompleteAttributeCorrectlyCheck());
+
+    checkMessagesVerifier.verify(sourceCode.getIssues())
+      .next().atLine(1).withMessage("DOM elements should use the \"autocomplete\" attribute correctly");
```
yassin-kammoun-sonarsource, line 84:
```suggestion
      .next().atLine(1).withMessage("DOM elements should use the \"autocomplete\" attribute correctly.");
```

sonar-html-plugin/src/test/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheckTest.java
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  @Test
+  void tel() {
+    HtmlSourceCode sourceCode = TestHelper.scan(new File("src/test/resources/checks/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck/tel.html"), new DomElementsShouldUseAutocompleteAttributeCorrectlyCheck());
+
+    checkMessagesVerifier.verify(sourceCode.getIssues())
+      .next().atLine(1).withMessage("DOM elements should use the \"autocomplete\" attribute correctly")
```
yassin-kammoun-sonarsource, line 92:
```suggestion
      .next().atLine(1).withMessage("DOM elements should use the \"autocomplete\" attribute correctly.")
```

sonar-html-plugin/src/test/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheckTest.java
```diff
@@ -0,0 +1,118 @@
...
+      .next().atLine(3)
+      .next().atLine(4);
+  }
+
+  @Test
+  void withContext() {
+    HtmlSourceCode sourceCode = TestHelper.scan(new File("src/test/resources/checks/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck/with-context.html"), new DomElementsShouldUseAutocompleteAttributeCorrectlyCheck());
+
+    checkMessagesVerifier.verify(sourceCode.getIssues())
+      .next().atLine(1).withMessage("DOM elements should use the \"autocomplete\" attribute correctly")
```
yassin-kammoun-sonarsource, line 103:
```suggestion
      .next().atLine(1).withMessage("DOM elements should use the \"autocomplete\" attribute correctly.")
```

sonar-html-plugin/src/test/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheckTest.java
```diff
@@ -0,0 +1,118 @@
...
+      .next().atLine(2);
+  }
+
+  @Test
+  void withWebauthn() {
+    HtmlSourceCode sourceCode = TestHelper.scan(new File("src/test/resources/checks/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck/with-webauthn.html"), new DomElementsShouldUseAutocompleteAttributeCorrectlyCheck());
+
+    checkMessagesVerifier.verify(sourceCode.getIssues())
+      .next().atLine(1).withMessage("DOM elements should use the \"autocomplete\" attribute correctly")
```
yassin-kammoun-sonarsource, line 112:
```suggestion
      .next().atLine(1).withMessage("DOM elements should use the \"autocomplete\" attribute correctly.")
```

yassin-kammoun-sonarsource: No need to change, but there is no need to assert the issue message if it is static. Checking it once should be enough.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck.java
```diff
@@ -0,0 +1,323 @@
...
+
+      return isValid;
+    }
+
+    int size() {
+      return constraints.size();
+    }
+  }
+
+  static class Token {
```
yassin-kammoun-sonarsource, line 74: Let's document what we mean by tokens with a sentence or two and an example.

```suggestion
  /** ... */
  static class Token {
```

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck.java
```diff
@@ -0,0 +1,323 @@
...
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+
+@Rule(key = "S6840")
+public class DomElementsShouldUseAutocompleteAttributeCorrectlyCheck extends AbstractPageCheck {
+  static boolean isADirective(String value) {
+    return value.startsWith("<%=") || value.startsWith("<?");
+  }
+
+  static class Validator {
```
yassin-kammoun-sonarsource, line 36: Let's document what kind of validation we are doing with a sentence or two and an example.

```suggestion
  /** ... */
  static class Validator {
```

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck.java
```diff
@@ -0,0 +1,323 @@
...
+    Token(
+      String value,
+      Predicate<TagNode> controlGroup
+    ) {
+      this.value = value;
+      this.controlGroupPredicate = controlGroup;
+    }
+  }
+
+  static class Candidate {
```
yassin-kammoun-sonarsource, line 88: Let's document what we mean by candidate with a sentence or two and an example.

```suggestion
  /** ... */
  static class Candidate {
```

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck.java
```diff
@@ -0,0 +1,323 @@
...
+    public final String value;
+
+    public final TagNode tagNode;
```
yassin-kammoun-sonarsource, line 91: Any reason why these fields are public while those from the Token class are private?

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/ControlGroup.java
```diff
@@ -0,0 +1,138 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.sonar.plugins.html.api.accessibility;
+
+import org.sonar.plugins.html.api.HtmlConstants;
+import org.sonar.plugins.html.node.TagNode;
+
+public class ControlGroup {
```
yassin-kammoun-sonarsource, line 23: Do you foresee using this class in a different context than the rule you implement in this PR?
I just wonder whether we should move it to the rule class.
I am not saying we should, just trying to understand :)

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/ControlGroup.java
```diff
@@ -0,0 +1,138 @@
...
+  public static boolean belongsToAutofillExpectationMantleControlGroup(TagNode node) {
+    if (!node.getNodeName().equals(HtmlConstants.NAME_INPUT)) {
+      return true;
+    }
+
+    String type = node.getAttribute("type");
```
yassin-kammoun-sonarsource, line 29: Let's use the `var` keyword as much as we can and let the Java compiler infer the type for us.

```suggestion
    var type = node.getAttribute("type");
```

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck.java
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    boolean isValid(TagNode tagNode) {
+      String autocomplete = tagNode.getAttribute("autocomplete");
+
+      if (autocomplete == null) {
+        return true;
+      }
+
+      String[] tokens = isADirective(autocomplete) ? new String[]{autocomplete} : autocomplete.split(" ");
```
yassin-kammoun-sonarsource, line 50:
```suggestion
      String[] tokens = isADirective(autocomplete) ? new String[]{autocomplete} : autocomplete.split("\\s+");
```

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/ControlGroup.java
yassin-kammoun-sonarsource: ... and let's do the same everywhere else in the code you introduced.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/ControlGroup.java
```diff
@@ -0,0 +1,138 @@
...
+
+    if (type != null && type.equals("tel")) {
+      return true;
+    }
+
+    return belongsToTextControlGroup(node);
+  }
+
+  public static boolean belongsToTextControlGroup(TagNode node) {
+    if (node.getNodeName().equals(HtmlConstants.NAME_TEXTAREA) || node.getNodeName().equals("select")) {
```
yassin-kammoun-sonarsource, line 99: Rule of three: let's extract `node.getNodeName()` into a variable.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/ControlGroup.java
```diff
@@ -0,0 +1,138 @@
...
+
+    if (type != null && type.equals("month")) {
+      return true;
+    }
+
+    return belongsToTextControlGroup(node);
+  }
+
+  public static boolean belongsToMultilineControlGroup(TagNode node) {
+    if (node.getNodeName().equals(HtmlConstants.NAME_TEXTAREA) || node.getNodeName().equals("select")) {
```
yassin-kammoun-sonarsource, line 55: Rule of three: let's extract `node.getNodeName()` into a variable.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/ControlGroup.java
```diff
@@ -0,0 +1,138 @@
...
+
+    return belongsToTextControlGroup(node);
+  }
+
+  public static boolean belongsToUsernameControlGroup(TagNode node) {
+    String type = node.getAttribute("type");
+
+    if (type != null && type.equals("email")) {
+      return true;
+    }
```
yassin-kammoun-sonarsource, line 131: After seeing this pattern multiple times, I think it would make sense to extract it into a method to `TagNode`:

```java
class TagNode {
  // ...
  public boolean hasAttributeValues(String attributeName, String... values) {
    var attributeValue = getAttribute(attributeName);
    if (attributeValue != null) {
      for (var value : values) {
        if (attributeValue.equalsIgnoreCase(value)) {
          return true;
        }
      }
    }
    return false;
  }
}
```

To be done in a different PR and (maybe go over all the rules that could benefit from this method.)

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck.java
```diff
@@ -0,0 +1,323 @@
...
+import org.sonar.plugins.html.checks.AbstractPageCheck;
+import org.sonar.plugins.html.node.TagNode;
+
+import java.util.List;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+
+@Rule(key = "S6840")
+public class DomElementsShouldUseAutocompleteAttributeCorrectlyCheck extends AbstractPageCheck {
+  static boolean isADirective(String value) {
```
yassin-kammoun-sonarsource, line 32: Let's move this static method below with all the other ones after all the inner classes you introduced.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/ControlGroup.java
```diff
@@ -0,0 +1,138 @@
...
+ * limitations under the License.
+ */
+package org.sonar.plugins.html.api.accessibility;
+
+import org.sonar.plugins.html.api.HtmlConstants;
+import org.sonar.plugins.html.node.TagNode;
+
+public class ControlGroup {
+  public static boolean belongsToAutofillExpectationMantleControlGroup(TagNode node) {
+    if (!node.getNodeName().equals(HtmlConstants.NAME_INPUT)) {
```
yassin-kammoun-sonarsource, line 25: Let's try to use `equalsIgnoreCase` instead of `equals` here and everywhere else.

```suggestion
    if (!node.getNodeName().equalsIgnoreCase(HtmlConstants.NAME_INPUT)) {
```

yassin-kammoun-sonarsource (approved):
I left minor suggestions here and there for you to address, but you are free to ignore any of them that you think are not worth the trouble.

Outstanding job!

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/ControlGroup.java
ericmorand-sonarsource: Definitely!

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck.java
ericmorand-sonarsource: Will do, and add a matching test.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck.java
ericmorand-sonarsource: I actually wondered if there was a naming convention for rules - `SXXXXCheck`, a sentence (like here), something descriptive, something imperative...

I'll apply your suggestion, and it would be cool if we could agree on a convention for future additions.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/ControlGroup.java
ericmorand-sonarsource: My main reasoning was that the rule class was getting very big. I'm not sure though this API makes sense outside of the autocomplete attribute - from my investigation in HTML spec there is no other mention of the concept of "Control group" outside of the autocomplete attribute spec, so it will probably never be used anywhere else...except if we have another rule at some point about this attribute.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/HtmlConstants.java
```diff
@@ -36,6 +36,13 @@ public class HtmlConstants {
   public static final String FILE_EXTENSIONS_DEF_VALUE = ".html,.xhtml,.cshtml,.vbhtml,.aspx,.ascx,.rhtml,.erb,.shtm,.shtml,.cmp,.twig";
   public static final String JSP_FILE_EXTENSIONS_PROP_KEY = "sonar.jsp.file.suffixes";
   public static final String JSP_FILE_EXTENSIONS_DEF_VALUE = ".jsp,.jspf,.jspx";
+
+  public static final String NAME_INPUT = "input";
+
+  public static final String NAME_TEXTAREA = "textarea";
+
+  public static final String TYPE_HIDDEN = "hidden";
```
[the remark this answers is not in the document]
yassin-kammoun-sonarsource: I think you can just inline the constants and mark the reported issues as won't fix. Part of the review process also includes reviewing the issues that are reported by SonarQube that were marked as "false positive" or "won't fix" by the author of the pull request.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck.java
yassin-kammoun-sonarsource: I don't think we have any convention, maybe we should. In general, we don't pay much attention to the check names since they are internal, but in your case, it was hard to miss it ^^

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/ControlGroup.java
yassin-kammoun-sonarsource: I see. I don't have a strong opinion on the matter so I let you decide whether you want to leave it there or move it.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/ControlGroup.java
ericmorand-sonarsource: True. Will fix.

Shouldn't this issue have been detected by SonarQube analysis?

sonar-html-plugin/src/test/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheckTest.java
ericmorand-sonarsource: Good point!

I will add the dot.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/HtmlConstants.java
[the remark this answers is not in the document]
ericmorand-sonarsource: Will do, thanks.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck.java
ericmorand-sonarsource: No reason, just an oversight. I'll streamline.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck.java
ericmorand-sonarsource: Fair point. I'll also rename the `constraints` property to the more semantically correct `predicates`.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/DomElementsShouldUseAutocompleteAttributeCorrectlyCheck.java
ericmorand-sonarsource: Good catch! I'll remove it entirely.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/ControlGroup.java
ericmorand-sonarsource: I'll keep it this way for now if that's OK for you.

yassin-kammoun-sonarsource: 👍

yassin-kammoun-sonarsource (approved):
LGTM!

PR #284, opened 2024-02-22
SONARHTML-198 Create rule S6852: Elements with an interactive role should support focus

yassin-kammoun-sonarsource:


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/FocusableInteractiveElementsCheck.java
```diff
@@ -0,0 +1,102 @@
...
+    return NON_INTERACTIVE_ELEMENTS.stream().anyMatch(tag -> tag.equalsIgnoreCase(element.getNodeName()));
+  }
+
+  private static boolean hasPresentationRole(TagNode element) {
+    return PRESENTATION_ROLES.stream().anyMatch(role -> role.equalsIgnoreCase(element.getAttribute("role")));
+  }
+
+  private static boolean hasNoninteractiveRole(TagNode element) {
+    return NON_INTERACTIVE_ROLES.stream().anyMatch(role -> role.equalsIgnoreCase(element.getAttribute("role")));
+  }
```
ilia-kebets-sonarsource, line 101: could you refactor this to use the methods from `HtmlConstants.java`, and move the new ones there?

ilia-kebets-sonarsource (approved):
Small refactoring suggestion, otherwise LGTM

PR #291, opened 2024-02-27
Fix S6825: align with React implementation

ilia-kebets-sonarsource:


ericmorand-sonarsource (approved):

mshima:
Looks like we have a few more false positives not related to this one: https://github.com/jhipster/generator-jhipster/issues/26106#issuecomment-2105128221.

Is there a public repository where we can fill a bug report?

ericmorand-sonarsource:
Hi @mshima, I've been looking at some of the lines that trigger the rule, like this [one](https://sonarcloud.io/project/issues?issues=AY8KoVrlXjJFDT-byu2i&open=AY8KoVrlXjJFDT-byu2i&id=jhipster-sample-application), and your [comment](https://github.com/jhipster/generator-jhipster/issues/26106#issuecomment-2105128221) on the mentioned `jhipster` issue:

> Doesn't make sense to require keypress on aria-hidden elements that are not focusable.

I feel that the rule should not trigger because, as you say, the element is not focusable. However, I'm also not sure how the rendered HTML page would work with assistive technology: since the close action is not reachable using the keyboard (both because the element is not focusable and because of the `aria-hidden` attribute), how can assistive technology users close the modal?

I want to ensure that this FP doesn't fortuitously detect a pattern that may impact impaired users negatively.

Note that the mentioned line of code should trigger [S6848](https://sonarsource.github.io/rspec/#/rspec/S6848) (_Non-interactive DOM elements should not have an interactive handler_).

mshima:
@ericmorand-sonarsource thanks for the response.

> I feel that the rule should not trigger because, as you say, the element is not focusable. However, I'm also not sure how the rendered HTML page would work with assistive technology: since the close action is not reachable using the keyboard (both because the element is not focusable and because of the `aria-hidden` attribute), how can assistive technology users close the modal?

There is a cancel button below, so this other close 'button' is visual only.

> I want to ensure that this FP doesn't fortuitously detect a pattern that may impact impaired users negatively.

Agreed.

> Note that the mentioned line of code should trigger [S6848](https://sonarsource.github.io/rspec/#/rspec/S6848) (_Non-interactive DOM elements should not have an interactive handler_).

That's valid.

It's a legacy code, Web:MouseEventWithoutKeyboardEquivalentCheck is a false positive in this case and doesn't justifies adding a keypress event.
S6848 is indeed an issue, that we should decide if we remove the visual only button or ignore it.

PR #293, opened 2024-03-19
Fix S6793 for aria-haspopup property

dougwaldron:
This PR fixes a false-positive for [rule S6793](https://sonarsource.github.io/rspec/#/rspec/S6793/html): The HTML plugin returns a false-positive for the ARIA property `aria-haspopup`. 

The [`aria-haspopup` property](https://w3c.github.io/aria/#aria-haspopup) allows the following token values: "false, true, menu, listbox, tree, grid, dialog". But if any value other than true or false is used, the following error is reported: *The value of the attribute "aria-haspopup" must be a boolean.*

See [this post on Sonar Community](https://community.sonarsource.com/t/fp-s6793-aria-property-aria-haspopup-allows-token-values-but-is-treated-as-a-boolean/111606) for background.


ilia-kebets-sonarsource:
fixed in https://github.com/SonarSource/sonar-html/pull/294

PR #298, opened 2024-04-08
Simplify file line check

zglicz:
Noticed that there was some overkill included in this rule. Let us simplify.


vdiez (approved):
good catch!

PR #300, opened 2024-04-09
Add rule S6811: DOM elements with ARIA role should only have supported properties

vdiez:
Fixes https://sonarsource.atlassian.net/browse/SONARHTML-208


sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/Property.java
```diff
@@ -0,0 +1,96 @@
...
+ */
+
+package org.sonar.plugins.html.api.accessibility;
+
+import java.util.Arrays;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+public enum Property {
```
zglicz, line 26: I'd prefer to call this AriaProperty. Property is a bit too generic.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/Role.java
```diff
@@ -0,0 +1,180 @@
...
+ */
+
+package org.sonar.plugins.html.api.accessibility;
+
+import java.util.Arrays;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+public enum Role {
```
zglicz, line 26: This could be as well coined as AriaRole

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/RoleSupportsAriaPropertyCheck.java
```diff
@@ -0,0 +1,64 @@
...
+import org.sonar.plugins.html.api.accessibility.Property;
+import org.sonar.plugins.html.api.accessibility.Role;
+import org.sonar.plugins.html.checks.AbstractPageCheck;
+import org.sonar.plugins.html.node.TagNode;
+
+@Rule(key = "S6811")
+public class RoleSupportsAriaPropertyCheck extends AbstractPageCheck {
+  @Override
+  public void startElement(TagNode element) {
+    var roleAttr = element.getAttribute("role");
```
zglicz, line 32: do you use getAttribute or getProperty. As I understand getProperty better handles how angular or vue works. At least looking at the angular documentation, they can pass them using their weird style and getProperty solves that for you

zglicz (approved):
lgtmacro

Although check the minor comments

PR #302, opened 2024-04-10
Add s6821

zglicz:
Mimicking logic from: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/rules/aria-role.js

Not including the options for now, as there also aren't any defaults in the original implementation.

The 1 thing I worry is about the templating languages hence using getAttribute vs getProperty to find mostly vanilla calls of : <... rule=""> instead of the vue or angular style attributes, where they can pass variables.

### DIFF
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/HtmlConstants.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/HtmlConstants.java
@@ -19,6 +19,8 @@ package org.sonar.plugins.html.api;
 
 import java.util.List;
 import java.util.Set;
+
+import org.sonar.plugins.html.api.accessibility.AriaRole;
 import org.sonar.plugins.html.node.TagNode;
 
 public class HtmlConstants {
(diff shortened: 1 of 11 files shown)


sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/HtmlConstants.java
```diff
@@ -243,7 +246,7 @@ public static boolean hasPresentationRole(TagNode element) {
 
   public static boolean hasAbstractRole(TagNode element) {
     var role = element.getAttribute("role");
-    return role != null && ABSTRACT_ROLES.stream().anyMatch(role::equalsIgnoreCase);
+    return role != null && AriaRole.of(role) != null && ABSTRACT_ROLES.stream().anyMatch(AriaRole.of(role)::equals);
```
vdiez, line 250:
```suggestion
    return role != null && AriaRole.of(role) != null && ABSTRACT_ROLES.stream().anyMatch(AriaRole.of(role.toLowerCase(Locale.ROOT))::equals);
```

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/AriaRoleCheck.java
```diff
@@ -0,0 +1,45 @@
...
+
+@Rule(key = "S6821")
+public class AriaRoleCheck extends AbstractPageCheck {
+    @Override
+    public void startElement(TagNode element) {
+        var role = element.getAttribute("role");
+        if (role == null) {
+            return;
+        }
+        var values = role.split(" ");
```
vdiez: can the role attribute have multiple values? if so, we should revisit implementation of other rules

vdiez: @ericmorand-sonarsource thoughts?

vdiez (approved):
just some minor comments

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/AriaRoleCheck.java
zglicz: seems that way - https://www.w3.org/TR/2014/REC-wai-aria-20140320/host_languages#host_general_role

```The attribute [value](https://www.w3.org/TR/2014/REC-wai-aria-20140320/terms#def_value) MUST allow a token list as the value;```

ericmorand-sonarsource: Yes, this is confirmed by the specification of the attribute itself: https://www.w3.org/TR/role-attribute/#s_role_module_attributes

> The role attribute takes as its value one or more whitespace separated TERMorCURIEorAbsIRIs, which is defined in [[RDFA-CORE](https://www.w3.org/TR/role-attribute/#bib-RDFA-CORE)]

PR #303, opened 2024-04-10
Implement S6822 - No redundant roles check

zglicz:
Porting logic from: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/rules/no-redundant-roles.js


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoRedundantRolesCheck.java
```diff
@@ -0,0 +1,93 @@
...
+    String[] pairs = allowedRedundantRoles.split(",");
+    var map = new HashMap<String, Set<AriaRole>>();
+    for (String pair : pairs) {
+      String[] parts = pair.split("=");
+      if (parts.length != 2) {
+        continue;
+      }
+      String tag = parts[0].trim();
+      String role = parts[1].trim();
+      map.computeIfAbsent(tag, k -> new HashSet<AriaRole>()).add(AriaRole.of(role));
```
vdiez, line 89: check if having multiple allowed roles works pls

vdiez (changes requested):

vdiez (approved):
LGTM!

PR #310, opened 2024-04-12
Resolve SONARHTML-217

ericmorand-sonarsource:


ericmorand-sonarsource:
As explained in the Jira issue, I left the summary behavior untouched - adding an exception for it was a deliberate move. We should probably have a more thorough discussion about this element in the future.

vdiez (approved):
LGTM!

PR #311, opened 2024-04-12
SONARHTML-230  Fix sonar quality issues

zglicz:


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/TableHeaderReferenceCheck.java
```diff
@@ -128,8 +128,7 @@ List<Set<String>> findHorizontalHeaders() {
         headers.add(new HashSet<>());
       }
       forEachCell((cell, row, column) -> {
-        if (cell instanceof Table.Header) {
-          Table.Header header = (Table.Header) cell;
+        if (cell instanceof Table.Header header) {
```
zglicz, line 131: this is pretty cool, although in JS/Ts and python, after an instanceof, you don't need to declare a new variable...

vdiez: yeah I also read about that when I saw the issues, very nice indeed!

vdiez (approved):
LGTM!

PR #313, opened 2024-04-15
SONARHTML-214  Handle dynamic values for S6821

zglicz:
Checked in peach and found a few issues with PHP inputting dynamic roles: https://peach.aws-prd.sonarsource.com/issues?projects=php%3Amagento2&resolved=false&rules=Web%3AS6821&open=AY7MwfK6A7MosuAUJtRS

### DIFF
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/AriaRoleCheck.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/AriaRoleCheck.java
@@ -20,6 +20,7 @@ package org.sonar.plugins.html.checks.accessibility;
 import static org.sonar.plugins.html.api.HtmlConstants.isAbstractRole;
 
 import org.sonar.check.Rule;
+import org.sonar.plugins.html.api.Helpers;
 import org.sonar.plugins.html.api.accessibility.AriaRole;
 import org.sonar.plugins.html.checks.AbstractPageCheck;
 import org.sonar.plugins.html.node.TagNode;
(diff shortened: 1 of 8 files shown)


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/AriaRoleCheck.java
```diff
@@ -29,7 +30,7 @@ public class AriaRoleCheck extends AbstractPageCheck {
   @Override
   public void startElement(TagNode element) {
     var role = element.getAttribute("role");
-    if (role == null) {
+    if (role == null || Helpers.isDynamicValue(role)) {
```
vdiez, line 33: Can you add this same check on all other rules checking `role == null` please?

vdiez (changes requested):
really nice, I think this should be a general check for all rules checking for role value

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/MouseEventWithoutKeyboardEquivalentCheck.java
```diff
@@ -42,6 +43,9 @@ public void startElement(TagNode node) {
       }
 
       var roleAttributeValue = node.getAttribute("role");
+      if (roleAttributeValue != null && Helpers.isDynamicValue(roleAttributeValue)) {
+        return;
+      }
```
zglicz, line 48: This one is special as in, it handles the null case with implicit roles.

vdiez (approved):

PR #361, opened 2025-07-09
Add license generation

zglicz:
Following the work from SonarJS maven - https://github.com/SonarSource/SonarJS/pull/5530

Very similar, needed again to copy over 2 licenses, as they tend to fail the downloads... :( probably some issues with the used maven plugin, as in JS this didn't happen once.

### DIFF
--- a/sonar-html-plugin/pom.xml
+++ b/sonar-html-plugin/pom.xml
@@ -182,4 +182,111 @@
       </plugin>
     </plugins>
   </build>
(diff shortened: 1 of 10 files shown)


vdiez (approved):
Looking good! btw, did you check if the generated jar actually contains the license files?

zglicz:
Yes, there is a folder `licenses/` there

PR #424, opened 2025-09-25
fix(Web:InternationalizationCheck): Ensure ignoredContentRegex is not bypassed

racetiger:
some HTML internationalization tools are based on source code annotation in HTML with special attributes (e.g `translate`).

For example: `<label for="username" translate>Username:</label>`

In that case `ignoredContentRegex` can be used to avoid triggering `Web:InternationalizationCheck` issues.

But the evaluation of the regexp is not independent from others checks on the same code block, and this can create false positives because the regexp is bypassed  :

```java
private boolean isValidText(String value) {
    return !isUnifiedExpression(value) && hasNoPunctuationOrSpace(value) && !isIgnoredByRegex(value);
``` 
In the example below we have set up `ignoredContentRegex` to ignore blocks annotated with `translate`.

The first two code blocks matching the `ignoredContentRegex` are still flagged with `Web:InternationalizationCheck` , while the third is properly ignored. This is because the hasNoPunctuationOrSpace is matched for the first two blocks, and the regexp check is bypassed.

InternationalizationCheck-issue

The proposed fix introduces a guard clause  in  `isValidText` to avoid this situation

### DIFF
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/coding/InternationalizationCheck.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/coding/InternationalizationCheck.java
@@ -18,6 +18,7 @@ package org.sonar.plugins.html.checks.coding;
 
 import java.util.List;
 import java.util.regex.Pattern;
+
 import org.sonar.check.Rule;
 import org.sonar.check.RuleProperty;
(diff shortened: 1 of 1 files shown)


vdiez:
LGTM! Thanks for contributing!

racetiger:
Awesome, thanks for merging !

PR #437, opened 2025-10-03
SONARHTML-316 Add lang validation for all text content

zglicz:
[SONARHTML-316](https://sonarsource.atlassian.net/browse/SONARHTML-316)

Augmenting the rule so that all text content will raise if the currently active lang property is invalid. Keeping the html check as is.

[SONARHTML-316]: https://sonarsource.atlassian.net/browse/SONARHTML-316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


zglicz:
This rule gets a bit noisy when you don't have any lang enabled. Perhaps we limit the number the reports on a single file? What would be a good choice?

vdiez (approved):

PR #605, opened 2026-03-11
Add RequiredAttributeTemplateCheck as a new template rule

zglicz:
## Summary
- Adds a new `RequiredAttributeTemplateCheck` template rule — users can create custom instances with their own `tag.attribute` configurations
- Leaves the existing `RequiredAttributeCheck` untouched so existing users are not affected
- Instantiates the new template in ruling tests with `img.alt` as a concrete example (`Template_RequiredImgAlt`)

> Note: the ruling expected file `Web-Template_RequiredImgAlt.json` is currently empty and will need to be updated with actual results after running the ITs.

Fixes https://community.sonarsource.com/t/track-missing-expected-attributes/179339

🤖 Generated with [Claude Code](https://claude.com/claude-code)


vdiez (changes requested):
by doing this you are effectively removing the rule. I think we cannot spot it because ruling does not have expected issues for RequiredAttributeCheck. Double check adding issues on ruling for this rule, I expect nothing will be raised. Have a look that for ruling we instantiate a template if we want the rule enabled: https://github.com/SonarSource/sonar-html/blob/0070a57965aa04f31b41197e973da62d44724c37/its/ruling/src/test/java/org/sonar/web/it/WebRulingTest.java#L67

I would suggest to create a new rule for this, so we don't affect existing users

vdiez (approved):

AndreasIgelCC:
Thank you for solving that @zglicz!

PR #618, opened 2026-04-14
SONARHTML-368 Apply monorepo grouping rule to maven updates

vdiez:
## Summary
- override inherited global grouping by using a monorepo-aware grouping rule
- group by sourceRepo when available, fallback to dependency name

## Managers
- applies to maven (and is future-proofed for 
pm)

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -12,6 +12,12 @@
   ],
   "packageRules": [
     {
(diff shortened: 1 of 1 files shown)


zglicz (approved):

vdiez:
Closing: we keep monorepo grouping for npm only, not maven.

PR #632, opened 2026-04-27
SONARHTML-373 sync rspec before 3.26 release and many WCAG updates

zglicz:
Part of


vdiez (approved):
sonar-html-plugin/src/main/resources/org/sonar/l10n/web/rules/Web/ParentElementIllegalCheck.html always points to wrong SQ replacement rule key. should use Web:ChildElementIllegalCheck instead of Web:S1931

PR #657, opened 2026-05-20
SONARHTML-146 Explicit method on form rule

erwan-leforestier-sonarsource:
**Summary**
  This PR adds `S8699`, a new HTML rule that raises an issue when a `<form>` tag does not declare an explicit valid `method` attribute.

  **Changes**
  - Implement the `S8699` analyzer rule and its rule metadata.
  - Accept `get`, `post`, and `dialog` as valid methods, case-insensitively.
  - Avoid false positives on supported dynamic/template syntaxes.

  **Functional Validation**
  An attached `sonarhtml-146-fv.zip` contains a bundled copy of `sonar-html`, a sample noncompliant HTML file, and a reproducer script.

  Run it from the extracted folder:
  1. Unzip [sonarhtml-146-fv.zip](https://github.com/user-attachments/files/28090689/sonarhtml-146-fv.zip)

  2. `cd sonarhtml-146-fv`
  3. `./run-reproducer.sh`

  Expected output:
  ```text
  Analyzed file: sample/noncompliant-form.html
  Issue count: 1
  Rule Web:S8699 at line 4: Use an explicit valid "method" attribute on this "<form>" tag ("get", "post", or "dialog").

### DIFF
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/attributes/FormMethodAttributeCheck.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/attributes/FormMethodAttributeCheck.java
@@ -68,46 +68,42 @@ public class FormMethodAttributeCheck extends AbstractPageCheck {
       return;
     }
     String thAttrValue = node.getAttribute("th:attr");
-    String thAttrMethodValue = thAttrValue == null ? null : extractThAttrMethodValue(thAttrValue);
-    if (thAttrMethodValue == null) {
+    if (thAttrValue == null) {
       createViolation(node, MESSAGE);
       return;
     }
-    if (isThymeleafExpression(thAttrMethodValue)) {
-      return;
-    }
-    if (!isValidMethod(thAttrMethodValue)) {
-      createViolation(node, MESSAGE);
-    }
-  }
-
-  private static String extractThAttrMethodValue(String thAttrValue) {
     Matcher matcher = TH_ATTR_METHOD_PATTERN.matcher(thAttrValue);
     if (!matcher.find()) {
-      return null;
+      createViolation(node, MESSAGE);
+      return;
     }
-    return unwrapSingleQuotes(matcher.group(1).trim());
+    String staticValue = staticThymeleafValue(matcher.group(1).trim());
+    if (staticValue == null) {
+      return;
+    }
+    if (!isValidMethod(staticValue)) {
+      createViolation(node, MESSAGE);
+    }
   }
 
-  private static String unwrapSingleQuotes(String value) {
+  private static String staticThymeleafValue(String value) {
+    if (value.startsWith("${") || value.startsWith("*{")
+      || value.startsWith("#{") || value.startsWith("@{")
+      || value.startsWith("~{")) {
+      return null;
+    }
     int len = value.length();
-    if (len >= 2 && value.charAt(0) == '\'' && value.charAt(len - 1) == '\'') {
+    if (len >= 2 && value.charAt(0) == '\'' && value.charAt(len - 1) == '\''
+      && value.indexOf('\'', 1) == len - 1) {
       return value.substring(1, len - 1);
     }
-    return value;
+    if (value.isEmpty() || PLAIN_TOKEN_PATTERN.matcher(value).matches()) {
+      return value;
+    }
+    return null;
   }
 
   private static boolean isValidMethod(String value) {
     return VALID_METHODS.contains(value.trim().toLowerCase(Locale.ROOT));
-  }
-
-  private static boolean isThymeleafExpression(String value) {
-    if (value.isEmpty()) {
-      return false;
-    }
-    return value.startsWith("${") || value.startsWith("*{")
-      || value.startsWith("#{") || value.startsWith("@{")
-      || value.startsWith("~{")
-      || !PLAIN_TOKEN_PATTERN.matcher(value).matches();
   }
 }
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/attributes/FormMethodAttributeCheck.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/attributes/FormMethodAttributeCheck.java
@@ -1,0 +1,61 @@
+/*
+ * SonarQube HTML
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * You can redistribute and/or modify this program under the terms of
+ * the Sonar Source-Available License Version 1, as published by SonarSource Sàrl.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+package org.sonar.plugins.html.checks.attributes;
+
+import java.util.Locale;
+import java.util.Set;
+import org.sonar.check.Rule;
+import org.sonar.plugins.html.api.Helpers;
+import org.sonar.plugins.html.checks.AbstractPageCheck;
+import org.sonar.plugins.html.node.Attribute;
+import org.sonar.plugins.html.node.TagNode;
+
+@Rule(key = "S8699")
+public class FormMethodAttributeCheck extends AbstractPageCheck {
+
+  private static final String METHOD_ATTRIBUTE = "method";
+  private static final Set<String> VALID_METHODS = Set.of("get", "post", "dialog");
+  private static final String MESSAGE = "Use an explicit valid \"method\" attribute on this \"<form>\" tag (\"get\", \"post\", or \"dialog\").";
+
+  @Override
+  public void startElement(TagNode node) {
+    if (!node.equalsElementName("form")) {
+      return;
+    }
+
+    Attribute methodAttribute = node.getProperty(METHOD_ATTRIBUTE);
+    if (methodAttribute == null) {
+      createViolation(node, MESSAGE);
+      return;
+    }
+
+    if (isDynamicMethod(methodAttribute)) {
+      return;
+    }
+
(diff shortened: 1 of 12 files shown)


vdiez (changes requested):
﻿Blocking issue in `sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/attributes/FormMethodAttributeCheck.java`: the Thymeleaf fast-path treats any `th:method` as compliant, even when the value is static and invalid.

With the current logic, `<form th:method="put">` reaches `hasThymeleafMethod(node)` and returns before validation, so no issue is raised. That contradicts the RSPEC note that only dynamic or framework-bound values that cannot be determined statically should be skipped.

I verified this locally by extending `FormMethodAttributeCheckTest` / `thymeleaf-invalid.html` with `th:method="put"`; the focused test then fails because no violation is reported. Please validate static Thymeleaf values the same way as plain `method` values and only skip dynamic ones.

vdiez (changes requested):
﻿Correction to my previous review: `th:method="put"` should not be raised here.

In Thymeleaf Spring, `th:method` is special-cased for non-browser methods: the rendered form is forced to `method="post"` and a hidden `_method` field is injected. Spring's `HiddenHttpMethodFilter` then maps `_method` to `PUT`/`DELETE`/`PATCH`, so `th:method="put"` is valid in this context.

The remaining issue is `th:attr="method=..."`. `th:attr` is just generic attribute assignment, so `th:attr="method=put"` would still render an invalid HTML form method. The current `hasThymeleafMethod` helper suppresses any `th:attr` containing `method=` without validating the static value, so that case becomes a false negative. Dynamic `th:attr="method=${...}"` should still be skipped, but static invalid values should raise.

erwan-leforestier-sonarsource:
@vdiez Thanks for the catch. Fixed in bb13f5cf:

- `th:method` is left as an unconditional skip, as you described — Spring's `HiddenHttpMethodFilter` makes any value valid there.
- For `th:attr`, the helper now extracts the value bound to `method=` and validates it: dynamic values (`${…}`, `*{…}`, `#{…}`, `@{…}`, `~{…}`) are skipped, static values are checked against `get/post/dialog`, so `th:attr="method=put"` and `th:attr="method=foo"` now raise.
- `thymeleaf-invalid.html` was extended with the static-invalid cases (`method=put`, `method=foo`, mixed with another attr, empty `method=`) and `thymeleaf.html` with `th:method="put"` plus the other Thymeleaf expression prefixes to pin the dynamic-skip behaviour.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/attributes/FormMethodAttributeCheck.java
```diff
@@ -0,0 +1,97 @@
...
+    if (!isValidMethod(methodAttribute.getValue())) {
+      createViolation(node, MESSAGE);
+    }
+  }
+
+  private boolean isDynamicMethod(Attribute methodAttribute) {
+    return !METHOD_ATTRIBUTE.equalsIgnoreCase(methodAttribute.getName())
+      || isUnifiedExpression(methodAttribute.getValue())
+      || Helpers.isDynamicValue(methodAttribute.getValue(), getHtmlSourceCode());
+  }
```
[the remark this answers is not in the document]
erwan-leforestier-sonarsource: Not applying. `Attribute.getValue()` defaults to `""` (see `Attribute(String name)` calling `this(name, "")`), so the null guard is unreachable in practice — SonarQube confirmed this by raising `java:S2589` on the analogous `value != null` in `isValidMethod`. Removed that check in 5cdf66e4 to match the same reasoning here. Per project convention we don't add defensive guards for cases that can't happen.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/attributes/FormMethodAttributeCheck.java
```diff
@@ -0,0 +1,97 @@
...
+    if (node.hasProperty("th:method")) {
+      return;
+    }
+    String thAttrValue = node.getAttribute("th:attr");
+    String thAttrMethodValue = thAttrValue == null ? null : extractThAttrMethodValue(thAttrValue);
+    if (thAttrMethodValue == null) {
+      createViolation(node, MESSAGE);
+      return;
+    }
+    if (isThymeleafExpression(thAttrMethodValue)) {
```
vdiez: This still reports some valid Thymeleaf `th:attr` assignments. `th:attr` accepts full Standard Expressions on the right-hand side, so `th:attr=method='post'` is valid, but `extractThAttrMethodValue()` returns `'post'` including the quotes, `isThymeleafExpression()` does not skip it, and `isValidMethod()` raises.

I reproduced this locally with a one-line scratch fixture under `target/`, and the check reported 1 issue. Please either recognize quoted / other non-static Thymeleaf expressions here or add the missing test coverage so this behavior is pinned.

vdiez (changes requested):
The earlier `th:attr=method=...` gap is fixed, but there is still one remaining false positive in the Thymeleaf path, so I’m keeping this blocked for now.

See the inline comment for the reproducible case.

Small non-blocking follow-up: `isValidMethod()` still carries a null guard even though parsed attribute values in this lexer are always strings.

vdiez:
Pulled the latest head: commit `5cdf66e4` addresses the small cleanup note from my review body by removing the dead null-check in `isValidMethod()`.

The blocking Thymeleaf comment still stands though, because the `th:attr` handling itself is unchanged in that commit.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/attributes/FormMethodAttributeCheck.java
erwan-leforestier-sonarsource: Valid catch — `'post'` is a Thymeleaf literal token, equivalent to the bare-token form we already accept. Fixed in 842929ad:

- `extractThAttrMethodValue` now unwraps balanced single quotes, so `'post'` → `post` and validates the same way as `post`.
- `isThymeleafExpression` now also returns `true` for anything that is not pure letters, so concatenations (`'po' + 'st'`), conditionals (`(c) ? 'a' : 'b'`), and operator forms in general are skipped instead of misclassified as invalid.
- Quoted invalid literals such as `'put'` still raise.

Fixtures pin the new cases (`method='post'`, `method='GET'`, `method=(c) ? 'post' : 'get'`, `method='po' + 'st'`, `method='put'`, `method='foo'`).

The `isValidMethod` null guard follow-up was already removed in 5cdf66e4 (SonarQube `java:S2589`).

A few cases consciously left alone — please flag if any matter:
- `data-th-method` / `data-th-attr` (HTML5 spelling): not handled in `ImgWithoutAltCheck` / `LangAttributeCheck` either — repo convention.
- `:method.sync` (deprecated Vue 2 modifier): `getProperty` does not expand it.
- `<form method="put" th:method="post">`: plain attribute is validated without knowing Thymeleaf will replace it.
- Numeric / boolean Thymeleaf literals (`method=42`, `method=true`): conservatively skipped as expressions; not realistic in practice.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/attributes/FormMethodAttributeCheck.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  private static boolean isThymeleafExpression(String value) {
+    if (value.isEmpty()) {
+      return false;
+    }
+    return value.startsWith("${") || value.startsWith("*{")
+      || value.startsWith("#{") || value.startsWith("@{")
+      || value.startsWith("~{")
+      || !PLAIN_TOKEN_PATTERN.matcher(value).matches();
```
vdiez: This new fallback is too broad and creates a false negative for static quoted invalid literals. After `unwrapSingleQuotes()`, a value like `th:attr=method='foo-bar'` becomes `foo-bar`; because that does not match `[A-Za-z]+`, `isThymeleafExpression()` now treats it as dynamic and skips reporting entirely.

I reproduced this locally on the current head: the analyzer reports `0` issues for `th:attr=method='foo-bar'`, even though the value is still statically invalid. Please narrow this branch so it only skips real Thymeleaf expressions, and add a fixture for the quoted-invalid case.

vdiez (changes requested):
The original Thymeleaf false positive is fixed, but the latest heuristic introduces a new false negative on the same code path.

See the inline comment for the reproducible case: static quoted invalid values such as `th:attr=method='foo-bar'` are now skipped as if they were dynamic.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/attributes/FormMethodAttributeCheck.java
erwan-leforestier-sonarsource: Good catch — fixed in b8d4e9c5.

The fallback no longer routes unwrapped quoted literals back through the `!PLAIN_TOKEN_PATTERN` branch. `staticThymeleafValue` now:

- detects a clean single-quoted literal (outer `'`/`'` with no internal `'`) and returns the unwrapped contents as a static value to validate directly;
- treats values with internal `'` (e.g. `'po' + 'st'`) or non-plain bare forms (e.g. `(c) ? 'a' : 'b'`) as dynamic;
- otherwise validates bare letter tokens as static.

So `th:attr="method='foo-bar'"` now reports an issue, while `'po' + 'st'` and `(condition) ? 'post' : 'get'` remain (correctly) skipped as dynamic. New fixture line in `thymeleaf-invalid.html` and the expectation in `FormMethodAttributeCheckTest.thymeleafInvalidValues` pin the behaviour.

vdiez (approved):
Latest changes address my remaining review comments.

I rechecked the focused `FormMethodAttributeCheckTest` and the previously discussed Thymeleaf cases on the current head, including quoted valid literals, conditional expressions, and quoted invalid literals such as `th:attr=method='foo-bar'`.

vdiez (approved):
Latest changes address my remaining review comments.

I rechecked the focused `FormMethodAttributeCheckTest` and the previously discussed Thymeleaf cases on the current head, including quoted valid literals, conditional expressions, and quoted invalid literals such as `th:attr=method='foo-bar'`.

erwan-leforestier-sonarsource:
Thanks for the review @vdiez 🙏 Merging now

PR #660, opened 2026-05-22
SONARHTML-159 Make ItemTagNotWithinContainerTagCheck template-aware

erwan-leforestier-sonarsource:
## Summary
`Web:ItemTagNotWithinContainerTagCheck` now suppresses false positives on `<li>` / `<dt>` tags whose container is supplied by a templating framework (HTML `<template>`, Angular `<ng-template>`, ASP.NET `<asp:Repeater>`, JSTL `<c:forEach>`, Thymeleaf `<th:block>`, JSP, …) instead of appearing inline in the source file.

## Changes
- Add reusable `Helpers.hasTemplateAncestor(TagNode)` — walks the parent chain for `<template>`, `<ng-template>` and any namespaced (`contains ':'`) tag.
- `ItemTagNotWithinContainerTagCheck.startElement` returns early when the helper matches.
- Update the rule HTML to document the new exceptions; add a dedicated unit test + fixture covering the five templating scopes.

## Functional Validation
Attached: `SONARHTML-159-fv.zip`

Unzip and run:

    ./run.sh

----
## Summary by Gitar

- **Razor improvements:**
  - Added logic to ignore literal braces in HTML body content within `@section` blocks.
  - Updated test fixtures and verifiers to account for literal brace handling in `razor-section.cshtml`.


erwan-leforestier-sonarsource:
Functional validation bundle: see attached `SONARHTML-159-fv.zip` below.

Unzip and run `./run.sh` — it compiles a small standalone reproducer against the bundled plugin jar and applies, for each fixture, both the legacy and the new template-aware ancestor logic. Expected output is in `expected-output.txt`.

| Fixture | Legacy | Template-aware |
| --- | --- | --- |
| asp-repeater.html | 1 FP | 0 |
| html-template.html | 2 FP | 0 |
| ng-template.html | 1 FP | 0 |
| jstl-foreach.html | 1 FP | 0 |
| thymeleaf-each.html | 1 FP | 0 |
| plain-real-bug.html | 2 | 2 (sanity: real bugs still raised) |

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/Helpers.java
```diff
@@ -41,4 +41,39 @@ public static boolean isDynamicValue(String value, HtmlSourceCode code) {
...
+  }
+
+  private static boolean isTemplateLikeTag(TagNode node) {
+    String name = node.getNodeName();
+    if (name == null || name.isEmpty()) {
+      return false;
+    }
+    return "template".equalsIgnoreCase(name)
+      || "ng-template".equalsIgnoreCase(name)
+      || name.indexOf(':') >= 0;
```
vdiez: `name.indexOf(':') >= 0` is too broad here. It makes every namespaced ancestor act like an opaque templating scope, but tags such as `c:if`, `th:block`, and `jsp:include` do not provide `<ul>` / `<dl>` containers on their own. On this branch, `<c:if test="${show}"><li>Bad</li></c:if>` now produces 0 issues, so this turns real violations into false negatives.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/ItemTagNotWithinContainerTagCheck.java
```diff
@@ -25,6 +26,9 @@ public class ItemTagNotWithinContainerTagCheck extends AbstractPageCheck {
 
   @Override
   public void startElement(TagNode node) {
+    if (Helpers.hasTemplateAncestor(node)) {
```
vdiez: This only reasons over parsed tag ancestors, so it does not address the Razor case from SONARHTML-159. In this analyzer Razor sectioning lives in text nodes / another file, not in the DOM ancestor chain, so the Jira example with `@section tabs { <li>... }` plus `@RenderSection("tabs")` still raises on this branch. We need either a regression test for that scenario and handling for it, or to split Razor out of this ticket.

vdiez (changes requested):
I checked this against SONARHTML-159 and I still see two blocking issues.

The new `:`-based ancestor shortcut is too broad and suppresses genuine violations for namespaced tags that do not synthesize `<ul>` / `<dl>` containers. I validated that `<c:if test="${show}"><li>Bad</li></c:if>` now reports 0 issues on this branch.

The Jira ticket's Razor `@section` / `@RenderSection(...)` case is still not handled. This implementation only inspects parsed tag ancestors, and Razor sectioning is not represented that way here. I validated the Jira example on this branch and the `<li>` still raises.

Requesting changes until the suppression is narrowed and the Razor scenario is either covered here or explicitly split out.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/Helpers.java
```diff
@@ -41,4 +41,39 @@ public static boolean isDynamicValue(String value, HtmlSourceCode code) {
...
+  public static boolean hasTemplateAncestor(TagNode node) {
+    TagNode parent = node.getParent();
+    while (parent != null) {
+      if (isTemplateLikeTag(parent)) {
+        return true;
+      }
+      parent = parent.getParent();
+    }
+    return false;
+  }
```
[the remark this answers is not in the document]
erwan-leforestier-sonarsource: Agreed — this PR added a third copy of the walk, so worth centralizing now before it spreads further. Going with your suggestion: a generic `Helpers.hasAncestorMatching(TagNode, Predicate<TagNode>)`, with `hasTemplateAncestor` becoming a one-liner and the two private methods in this check refactored to use it too. Coming in the next commit.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/Helpers.java
erwan-leforestier-sonarsource: Fair push — `<c:if>` / `<c:when>` / `<th:block>` / `<jsp:include>` are control-flow or no-op wrappers and don't supply a container, so suppressing on them does hide real bugs. The bare-`:` rule was chasing the variety of FP patterns in the Jira ticket and in the ruling delta (ASP.NET WebForms, JSP custom taglibs, Thymeleaf, JSTL), but you're right that the FN risk is concrete and worth more than the FP coverage.

Before narrowing, the trade-off I want to call out: dropping the generic `:` match means JSP custom taglibs like Silverpeas `<view:window>` / `<view:tabs>` (4 entries in this PR's ruling delta) re-FP. Looking at one of them:

```jsp
<view:window>
    <view:tabs>...</view:tabs>
    <li><view:board>...
```

There's no `<ul>` anywhere in that snippet, so this is almost certainly a real bug masquerading as an FP. Same intuition for the `<c:when><dt>` / `<c:otherwise><dt>` rows in the Silverpeas `portletSP.jsp` — they only suppress if a `<dl>` exists upstream in the file (which the legacy ancestor walk would already accept). So narrowing is the right call.

Going with: `template`, `ng-template`, and the `asp:` prefix only. ASP.NET WebForms `<asp:Repeater>` / `<asp:DataList>` / `<asp:GridView>` provably wrap their `ItemTemplate` content with HeaderTemplate/FooterTemplate, which is the clear container-provision case. Everything else with a colon either doesn't supply a container or the ancestor walk already catches it via the legacy path.

Also updating the fixtures — the `<c:forEach>` / `<th:block>` / `<jsp:include>` rows in `valid-cases.html` were misclassified for the same reason you spotted, and need to move out (or be reframed as `<ul><c:forEach><li/></c:forEach></ul>` to test the wrapped case).

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/ItemTagNotWithinContainerTagCheck.java
erwan-leforestier-sonarsource: You're right — `@section` blocks are parsed as TextNodes in this analyzer (CLAUDE.md calls this out: `@` isn't tokenized as expressions like JSP `<% %>` is), so the `<li>` inside a Razor section ends up as a root-level TagNode with no parent. Pure ancestor walking can't reach it.

The pre-analysis I worked from listed `Razor/ERB text heuristics` as part of the pragmatic scope, with `UnclosedTagCheck.java:50-55` cited as the precedent — that rule disables itself on `.cshtml` entirely. I'd rather not go that far here (we'd lose real-bug detection on the well-formed `.cshtml` files where someone actually misplaces `<li>` inside `<div>`), so going for a surgical version: in `.cshtml` / `.vbhtml`, suppress only when the `<li>`/`<dt>` is at the document root (no tag ancestors at all). That's the shape of a Razor section partial — items meant for an outer layout's `<ul>` / `<dl>`.

What that covers:
- ✅ `@section tabs { <li>Tab heading</li> }` — `<li>` at root → suppressed.
- ✅ A standalone `.cshtml` partial that only contains `<li>` rows.
- ✅ Real bug `<div><li>misplaced</li></div>` in `.cshtml` — `<li>` has a `<div>` ancestor, so still flagged.

Adding a regression test under `ItemTagNotWithinContainerTagCheck/razor-section.cshtml`. Will not extend to `.erb` / `.rhtml` for this PR — Ruby ERB hasn't shown up in the ruling deltas, and I'd rather not broaden the surface area for a case we don't have evidence of yet. Happy to revisit if you'd prefer to bundle ERB in here.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/Helpers.java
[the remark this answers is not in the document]
erwan-leforestier-sonarsource: Resolved by the narrowing discussed in @vdiez's thread above — `isTemplateLikeTag` is becoming `template` / `ng-template` / `asp:`-prefixed only, so `<svg:rect>` / `<math:mi>` no longer match.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/ItemTagNotWithinContainerTagCheck.java
```diff
@@ -25,37 +26,41 @@ public class ItemTagNotWithinContainerTagCheck extends AbstractPageCheck {
 
   @Override
   public void startElement(TagNode node) {
+    if (Helpers.hasTemplateAncestor(node) || isRazorPartialRoot(node)) {
```
vdiez: `isRazorPartialRoot()` is broader than the Jira case: it suppresses every root-level `<li>` / `<dt>` in `.cshtml` / `.vbhtml`, not just `@section` content. I validated this with a plain Razor file containing only a top-level `<li>` and `<dt>`; this branch reports 0 issues. That fixes the reported FP by effectively disabling the rule for all root-level Razor items.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/Helpers.java
```diff
@@ -41,4 +42,60 @@ public static boolean isDynamicValue(String value, HtmlSourceCode code) {
...
+
+  private static boolean isTemplateLikeTag(TagNode node) {
+    String name = node.getNodeName();
+    if (name == null || name.isEmpty()) {
+      return false;
+    }
+    return "template".equalsIgnoreCase(name)
+      || "ng-template".equalsIgnoreCase(name)
+      || startsWithIgnoreCase(name, "asp:");
```
vdiez, line 82: `startsWithIgnoreCase(name, "asp:")` still makes every WebForms control template-like, but the ticket only justifies repeater/data-list style controls that actually wrap item content with generated list markup. Wrappers such as `asp:PlaceHolder` do not provide `<ul>` / `<dl>` containers; `<asp:PlaceHolder runat="server"><li>x</li></asp:PlaceHolder>` now reports 0 issues on this branch.

vdiez (changes requested):
I re-reviewed the latest update and there are still two blocking regressions.

The Razor handling now suppresses every root-level `<li>` / `<dt>` in `.cshtml` / `.vbhtml`, not just section content. I validated that a file containing only a top-level `<li>` and `<dt>` now reports 0 issues on this branch.

The ASP.NET heuristic is also still broader than the cases justified by the ticket. Treating every `asp:*` ancestor as template-like hides genuine violations for wrappers that do not synthesize list containers; I validated that `<asp:PlaceHolder runat="server"><li>x</li></asp:PlaceHolder>` also reports 0 issues.

Requesting changes until both heuristics are narrowed to the specific templating cases they are meant to cover.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/Helpers.java
erwan-leforestier-sonarsource: I'v been asking codex about all the possible repeater/data-list style controls and here is the answer:

```
A .aspx file mixes plain HTML with server controls (tags like <asp:Something runat="server">). At request time, the server expands those server controls into final HTML and sends
  that to the browser. So what the developer writes is not what the browser sees — and that gap is what trips up the HTML analyzer.

  The analyzer reads the .aspx file as if it were HTML and sees a <li> floating inside a <asp:Repeater> tag, which doesn't look like a <ul>/<ol>/<dl>. False positive.

  Now the three concrete cases:

  ---
  1. asp:Repeater — header/footer hold the wrapper

  The developer writes one item's markup once. The Repeater stamps it out N times.

  What the developer writes (.aspx):
  <asp:Repeater ID="Fruits" runat="server">
      <HeaderTemplate>
          <ul>
      </HeaderTemplate>
      <ItemTemplate>
          <li><%# Eval("Name") %></li>
      </ItemTemplate>
      <FooterTemplate>
          </ul>
      </FooterTemplate>
  </asp:Repeater>

  What the browser receives (assuming 3 data rows):
  <ul>
      <li>Apple</li>
      <li>Banana</li>
      <li>Cherry</li>
  </ul>

  Why the analyzer is confused: in the source file, <ul> and <li> are inside sibling template elements, not in an ancestor/descendant relationship. The opening <ul> is even
  unbalanced (no closing tag in the same template) — HeaderTemplate and FooterTemplate together produce balanced output, but each one on its own looks broken.

  ---
  2. asp:ListView — wrapper lives in LayoutTemplate, items get spliced in

  ListView is the more modern version of Repeater. Instead of a header/footer pair, it has one LayoutTemplate containing a placeholder that marks where items go.

  What the developer writes (.aspx):
  <asp:ListView ID="Fruits" runat="server">
      <LayoutTemplate>
          <ul>
              <li runat="server" id="itemPlaceholder" />
          </ul>
      </LayoutTemplate>
      <ItemTemplate>
          <li><%# Eval("Name") %></li>
      </ItemTemplate>
  </asp:ListView>

  The element with id="itemPlaceholder" is a marker — at render time it's replaced by the concatenation of all ItemTemplate outputs.

  What the browser receives:
  <ul>
      <li>Apple</li>
      <li>Banana</li>
      <li>Cherry</li>
  </ul>

  Why the analyzer is confused: the <li> inside ItemTemplate has no <ul> ancestor in the source file — the <ul> is over in LayoutTemplate. Worse, the <li runat="server"
  id="itemPlaceholder" /> in LayoutTemplate is itself a <li> directly inside <ul>, but it's a placeholder, not a real list item — disappears at render time.

  A variant uses a generic <asp:PlaceHolder> instead:
  <LayoutTemplate>
      <ul>
          <asp:PlaceHolder runat="server" ID="itemPlaceholder" />
      </ul>
  </LayoutTemplate>
  Same idea, no fake <li> in the source.

  ---
  3. asp:DataList — depends on RepeatLayout

  DataList also has HeaderTemplate / ItemTemplate / FooterTemplate, but the wrapper it generates around items is controlled by the RepeatLayout attribute. Four values matter:

  3a. RepeatLayout="Table" (default) — generates a <table>. If you put <li> in ItemTemplate here, it's actually broken HTML, not a FP. Analyzer is right to complain.

  3b. RepeatLayout="Flow" — generates <span> wrappers; developer fully controls header/footer markup, same pattern as Repeater:

  <asp:DataList ID="Fruits" runat="server" RepeatLayout="Flow">
      <HeaderTemplate><ul></HeaderTemplate>
      <ItemTemplate><li><%# Eval("Name") %></li></ItemTemplate>
      <FooterTemplate></ul></FooterTemplate>
  </asp:DataList>
  Renders the same <ul><li>...</li></ul> as Repeater. FP source.

  3c. RepeatLayout="UnorderedList" (added later) — the control itself emits the <ul>:

  <asp:DataList ID="Fruits" runat="server" RepeatLayout="UnorderedList">
      <ItemTemplate><%# Eval("Name") %></ItemTemplate>
  </asp:DataList>
  Renders:
  <ul>
      <li>Apple</li>
      <li>Banana</li>
      <li>Cherry</li>
  </ul>
  No raw <li> in the source file at all — the control wraps each item in <li> automatically. Not a FP source for this rule.

  3d. RepeatLayout="OrderedList" — same as 3c but emits <ol>.

  ---
  Secondary controls — not FP sources, but worth knowing

  - asp:BulletedList — emits the entire <ul>/<ol> and <li> structure itself; developer never writes raw <li>.
  - asp:RadioButtonList / asp:CheckBoxList with RepeatLayout="UnorderedList" — same, control emits the list markup.
  - asp:Menu with RenderingMode="List" — same.

  These don't trigger the rule because there's no raw <li> in the source for it to flag.

  ---
```

Imo, we should not be too framework specific and start handling all the possibilities that the framework offers. The goal here is to tackle down 90% of the FPs that were raised before with a wider heuristic. The counterparts of that is an increase in the rate of raised FNs but I think it's acceptable. 
Covering all the FPs without increasing the FNs is just impossible, for the simple reason that we could have reusable component files that contains top-level `<li>` that are then reused inside other components. The easiest way of not having FPs would be to completely ignore .aspx files but then the rate of FNs would explode.  
It's more a matter of finding the right balance between FPs and FNs. By just ignoring everything wrapped inside a `<asp:` tag, we would probably catch more of the FPs with a reasonable FNs rate increase.
Wdyt ?

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/ItemTagNotWithinContainerTagCheck.java
erwan-leforestier-sonarsource: It's true that it might be too wide. Narrowed it down to ignore only tags inside `@section`.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/ItemTagNotWithinContainerTagCheck.java
```diff
@@ -16,46 +16,132 @@
...
+    state.sectionStartLine = baseLine + countNewlines(code, 0, m.start());
+    state.depth = 1;
+    return m.end();
+  }
+
+  private int advanceBraceDepth(String code, int pos, int baseLine, ScanState state) {
+    char c = code.charAt(pos);
+    if (c == '{') {
+      state.depth++;
+    } else if (c == '}') {
```
vdiez, line 85: This counts any `{` / `}` in a `TextNode` as Razor structure. A literal brace in ordinary content, for example `<p>}</p>` inside an `@section`, closes the section early and makes later `<li>` / `<dt>` in the same section raise again. I reproduced that with a temporary local test, so I think the new Razor support needs to ignore literal braces in normal text before merge.

vdiez (changes requested):
I re-reviewed the latest head. I'm not re-blocking on the broader `asp:*` tradeoff, since the discussion now makes that an explicit FP/FN balance choice.

But I do see one remaining blocker in the new Razor `@section` scanner.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/ItemTagNotWithinContainerTagCheck.java
erwan-leforestier-sonarsource: @vdiez Addressed — pushed commit 1355559a.

**Fix:** `scanForSections` now early-returns when the TextNode has a parent. `TextNode.getParent()` is set by the lexer's `createNodeHierarchy` (and handles void elements correctly), so a non-null parent means the TextNode is HTML body content — its `{`/`}` are literal text, not Razor structure.

**Test:** added a `@section withLiteralBraces` case to `razor-section.cshtml` with `<p>literal close: }</p>` followed by a `<li>still inside @section</li>`. Reproduced the FP before the fix (got an issue on the `<li>` instead of the expected line), green after.

vdiez (approved):
Re-reviewed the latest head. The remaining Razor @section scanner issue is addressed, including coverage for literal braces inside normal HTML text. Looks good to me.

PR #661, opened 2026-05-22
SONARHTML-382 Fix SonarLint AnalysisWarnings injection

vdiez:
## Summary
- replace raw AnalysisWarnings injection in HtmlSensor with an AnalysisWarningsWrapper
- register the wrapper as a plugin extension so SonarQube gets delegated warnings while SonarLint can instantiate the no-op path
- update plugin and sensor tests for the new wiring

## Testing
- mvn -pl sonar-html-plugin -Dtest=HtmlPluginTest,HtmlSensorTest test


erwan-leforestier-sonarsource (approved):
Thanks for the quick fix @vdiez 🙏 One relevant comment from gitar but other than that LGTM. I've already approved it.

PR #662, opened 2026-05-22
SONARHTML-172 Handle S1093 in Angular templates

erwan-leforestier-sonarsource:
## Summary
S1093 reports a false positive when a list item is defined inside an Angular `ng-template` and rendered through `ngTemplateOutlet`. This change skips that check for Angular template definitions while preserving real detached-item findings.

## Changes
- Skip S1093 container validation for `li` and `dt` nodes declared under Angular `ng-template`
- Add a regression fixture covering the Angular `ngTemplateOutlet` case from the community report
- Keep detached list items in the new test fixture to prove valid findings still report

## Functional Validation
Attached: [SONARHTML-172-fv.zip](https://github.com/user-attachments/files/28154879/SONARHTML-172-fv.zip)

Unzip and run:
    ./run.sh

:warning::warning: This is not ready for review :warning::warning:


erwan-leforestier-sonarsource:
Functional validation bundle prepared as `SONARHTML-172-fv.zip`.

`gh` in this environment does not support uploading binary PR comment attachments, so the zip was generated locally at `/tmp/SONARHTML-172-fv.zip` and validated before drafting this PR.

Validated output:
```text
angular-template.html: 0 issue(s)
detached-items.html: 2 issue(s)
1: Surround this <li> item tag by a <ul> or <ol> container one.
2: Surround this <dt> item tag by a <dl> container one.
```

erwan-leforestier-sonarsource:
Closing — this is covered by #660 (SONARHTML-159), which already adds `Helpers.hasTemplateAncestor` that walks ancestors looking for `<ng-template>` (alongside HTML `<template>` and ASP.NET `asp:*` server controls). `ItemTagNotWithinContainerTagCheck.startElement` consults that helper before reporting, and PR #660's `valid-cases.html` fixture already exercises the Angular `<ng-template>` case. PR #660 is broader (also handles `<template>` / `asp:*`) and is the path forward.

PR #663, opened 2026-05-22
SONARHTML-170 Suppress S5256 false positives on Razor layout tables

erwan-leforestier-sonarsource:
## Summary
S5256 (*tables should have headers*) raises false positives on Razor layout pages, where the `<table>` sits in the layout and the `<th>` rows are injected by `@RenderBody`, `@RenderSection`, or `@Html.Partial`/`@Html.RenderPartial` from a child view. This PR adds a symmetric exemption to the existing Thymeleaf fragment-insertion logic.

## Changes
- `TableWithoutHeaderCheck.startDocument` pre-scans `.cshtml`/`.vbhtml` files for `TextNode`s containing Razor fragment-rendering expressions (`@RenderBody`, `@RenderSection`, `@RenderPage`, `@Html.(Render)?Partial(Async)?`, with or without `await`); every `<table>` ancestor of such a node is exempted from S5256.
- Reusable helpers added to `Helpers`: `isRazorFile(HtmlSourceCode)` and `containsRazorFragmentRendering(String)`.
- New unit tests + `.cshtml` / `.vbhtml` / `.html` fixtures covering the FP cases and the non-suppression invariants (real bug in a Razor file, Razor-looking text in a plain `.html` file).

## Functional Validation
Attached: [SONARHTML-170-fv.zip](https://github.com/user-attachments/files/28154579/SONARHTML-170-fv.zip)

Unzip and run:

    ./run.sh

### DIFF
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/Helpers.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/Helpers.java
@@ -23,6 +23,10 @@ import java.util.function.Predicate;
 import java.util.regex.Pattern;
 
 public class Helpers {
+
+  private static final Pattern RAZOR_FRAGMENT_RENDERING = Pattern.compile(
+    "@(?:RenderBody|RenderSection|RenderPage|(?:await\\s+)?Html\\.(?:Render)?Partial(?:Async)?)\\b");
+
   private Helpers() {
   }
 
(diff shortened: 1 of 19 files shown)


erwan-leforestier-sonarsource:
Functional validation bundle: see attached `[Uploading SONARHTML-170-fv.zip…]()below.

Unzip and run `./run.sh` — it compiles a small standalone reproducer against the bundled patched plugin JAR and applies, for each fixture, both the legacy and the new Razor-aware logic. Expected output is in `expected-output.txt`.

| Fixture | Legacy | Patched |
| --- | --- | --- |
| `_Layout.cshtml` | 1 FP | 0 |
| `_LayoutWithSection.cshtml` | 1 FP | 0 |
| `_PartialHost.cshtml` | 2 FP | 0 |
| `RealBug.cshtml` | 1 | 1 (sanity: real bug still raised) |
| `RazorTokensInPlainHtml.html` | 1 | 1 (sanity: literal text in `.html` not suppressed) |

vdiez (changes requested):
﻿I found one blocking issue in the new Razor suppression logic.

`containsRazorFragmentRendering` now treats any `@{ ... }` block containing `Html.Partial*` or `Component.Invoke*` as fragment rendering. That is too broad: those helpers return content, but they do not render anything when the result is assigned or otherwise discarded inside a code block.

Minimal repro:

```cshtml
<table>
  @{
    var rows = Html.Partial("_Rows");
  }
</table>
```

With the current branch, `S5256` is suppressed on that table even though no headers are rendered.

I verified this locally in the worktree by scanning a throwaway fixture with `TableWithoutHeaderCheck`; the analyzer reported `0` issues.

I think the `@{ ... }` path should stay limited to writer-style helpers such as `RenderPartial*` / `RenderAction`, while `Html.Partial*` / `Component.Invoke*` should only be considered in expression forms that actually emit output.

erwan-leforestier-sonarsource (commented):
Good catch — pushed 5247712e.

I narrowed `RAZOR_FRAGMENT_STATEMENT` to writer-style helpers only (`Html.RenderPartial(Async)?`, `Html.RenderAction`). `Html.Partial*` and `Component.Invoke*` now only count toward suppression in expression form (matched by `RAZOR_FRAGMENT_EXPRESSION`), which is unchanged.

Added a regression fixture `razor-code-block-assigned-result.cshtml` covering exactly your repro plus the `PartialAsync` and `Component.InvokeAsync` variants — all three tables now correctly raise S5256.

vdiez (approved):
Re-reviewed the latest head. The Razor suppression logic is now narrowed correctly, and the regression coverage for assigned partial/component results is in place. Looks good to me.

PR #664, opened 2026-05-22
SONARHTML-181 Align S1077 with JS rule

erwan-leforestier-sonarsource:
## Summary
This updates HTML rule S1077 so it accepts the same richer alternative-text inputs as the JS rule. It now handles ARIA-based alternatives and object elements instead of only alt-based checks.

## Changes
- Extend `ImgWithoutAltCheck` to accept non-empty `aria-label` and `aria-labelledby` values where applicable
- Add `<object>` handling to S1077 and cover the new cases with focused unit fixtures
- Regenerate the legacy-key HTML rule description from the updated RSPEC branch

## Functional Validation
Attached: `SONARHTML-181-fv.zip`

Unzip and run:
    ./run.sh

----
## Summary by Gitar

- **Refactored logic:**
  - Centralized validation in `requiresAlternativeText` to unify `img`, `input type="image"`, and `area` checks.
  - Added robust `th:attr` parsing to handle nested braces/parentheses and properly validate attribute assignments.
- **Improved test coverage:**
  - Added `ImgWithoutAltCheckAria.html` to verify `aria-label` and `aria-labelledby` compliance.
  - Added `ImgWithoutAltCheckThymeleaf.html` to validate empty/blank value rejection in complex Thymeleaf assignments.


erwan-leforestier-sonarsource:
Functional validation artifact: `SONARHTML-181-fv.zip`

Download:
https://raw.githubusercontent.com/SonarSource/sonar-html/artifacts/SONARHTML-181-fv/SONARHTML-181-fv.zip

vdiez (approved):
﻿Looks good to me.

The scope is clean, the rule logic is aligned with the SonarJS behavior we wanted, and the added coverage for ARIA-based alternatives plus Thymeleaf cases is in the right place.

One non-blocking follow-up worth considering: `hasNonEmptyAttribute` currently lets a non-empty static `alt` / `aria-*` mask an empty Thymeleaf override such as `th:alt=""` or `th:attr="aria-label=''"`. I would treat that as a nice fix or follow-up test rather than a blocker for this PR.

vdiez:
﻿Before merging, please make sure the changes from SonarSource/rspec#6908 have been pulled into this PR.

PR #667, opened 2026-05-26
SONARHTML-184 Fix incorrect S1458 tag reporting

erwan-leforestier-sonarsource:
## Summary
Fix S1458 so unopened closing tags do not cause the rule to blame the current open tag. Preserve the existing report when there is a real unclosed inner tag.

## Changes
- keep the open-tag stack intact when an end tag has no matching opener
- continue reporting the innermost unclosed tag when a later matching end tag exists
- add PHP and HTML regression tests covering the Jira reproducer and the preserved mismatch case

## Functional Validation
Attached: `SONARHTML-184-fv.zip`

Unzip and run:
    ./run.sh

----
## Summary by Gitar

- **Test improvements:**
  - Refactored multiple test cases into a single `ParameterizedTest` in `UnclosedTagCheckTest.java` to improve maintainability.


erwan-leforestier-sonarsource:
Functional validation bundle uploaded: [SONARHTML-184-fv.zip](https://github.com/SonarSource/sonar-html/releases/download/untagged-d818663a5db6b3904f3d/SONARHTML-184-fv.zip)

`expected-output.txt` is inside the zip; unzip and run `./run.sh` from the bundle.

vdiez (approved):
Re-reviewed the latest head. The stack handling now avoids blaming the current opener for stray closing tags while still reporting the actual unclosed inner tags. Looks good to me.

PR #668, opened 2026-05-26
SONARHTML-186 Narrow S1876 to interpolated HTML comments

erwan-leforestier-sonarsource:
## Summary
Narrows rule S1876 (`AvoidHtmlCommentCheck`) to only flag HTML comments that contain server-side interpolation, drastically cutting the SECURITY_HOTSPOT review noise on real projects while keeping the genuine information-leak signal.

## Changes
- `AvoidHtmlCommentCheck` now requires the comment to contain a dynamic marker (`${...}`, `<% ... %>`, `<%= ... %>`, `{{ ... }}`, `{% ... %}`, `<?= ... ?>`, `<?php ... ?>`, or Razor `@...`) in addition to being on a server-side page.
- Added `Helpers.containsDynamicValue` next to `isDynamicValue`, sharing the same marker set so the patterns stay in one place.
- Regenerated `AvoidHtmlCommentCheck.html` from the paired rspec branch (`rule/modify-RSPEC-S1876-narrow-detection`), and updated test fixtures to cover both interpolated and non-interpolated comments.

## Functional Validation
Attached: `SONARHTML-186-fv.zip`

Unzip and run:
    ./run.sh


erwan-leforestier-sonarsource:
Functional validation zip lives at `/tmp/SONARHTML-186-fv.zip` in the agent workspace. Please drag it into this PR's web UI to attach the artifact, or grab it from env-4 before releasing the workspace.

Inside the zip, `./run.sh` auto-detects a sonar-html checkout (or honors `SONAR_HTML_DIR`/first arg), builds the test classpath, compiles `Demo.java`, and prints the findings — expected output is in `expected-output.txt`.

vdiez (changes requested):
I pulled the latest head and checked it against SONARHTML-186. The ruling update is in the right direction, but I still see one blocker before this is ready.

AvoidHtmlCommentCheck.startDocument() still marks a file as server-side only when it sees a node starting with <?php or <%. That means several of the new interpolation forms added to Helpers.containsDynamicValue(...) are still missed whenever the comment is the only server-side marker in the file. I reproduced this locally with cases like  in a .jsp file and  in a .php file: both produce  issues unless some unrelated <?php ... ?> or <% ... %> appears elsewhere first.

The added tests currently hide that gap because they all include an extra legacy marker outside the comment:
- document-interpolation.jsp has <%-- ... --%>
- document.php has a separate <?php ... ?> block
- document.html.erb has <%= hello %> outside the comment

Also, the updated rule description says <%-- ... --%> is the server-side comment syntax for JSP/ERB. That is correct for JSP, but ERB comments use <%# ... %>.

Requesting changes because the ticket scope is specifically “HTML comments containing server-side interpolation”, and the current implementation still misses several of those cases.

vdiez:
Reposting the review as a normal PR comment because the earlier review body is not surfacing reliably on the PR page.

Main blocker:
`AvoidHtmlCommentCheck.startDocument()` still marks a file as server-side only when it sees a node starting with `<?php` or `<%`. Because of that, several of the newly supported interpolation forms are still missed when the HTML comment is the only server-side marker in the file.

I reproduced this locally with cases like:
- `<!-- ${user.name} -->` in a `.jsp` file
- `<!-- <?= $userId ?> -->` in a `.php` file

Both produce `0` issues unless some unrelated `<?php ... ?>` or `<% ... %>` appears elsewhere first.

The current added tests hide that gap because they all include an extra legacy marker outside the comment:
- `document-interpolation.jsp` has `<%-- ... --%>`
- `document.php` has a separate `<?php ... ?>` block
- `document.html.erb` has `<%= hello %>` outside the comment

Smaller point:
The updated rule description says `<%-- ... --%>` is the server-side comment syntax for JSP/ERB. That is correct for JSP, but ERB comments use `<%# ... %>`.

So the ruling update looks right, but I still think the implementation needs one more pass to align with `SONARHTML-186`'s stated scope: HTML comments containing server-side interpolation.

erwan-leforestier-sonarsource:
Both points addressed:

- **Gate**: `AvoidHtmlCommentCheck.startDocument()` now marks a file as server-side when the extension is a known template suffix (`.jsp`/`.jspf`/`.jspx`, `.php`/`.phtml`, `.cshtml`/`.vbhtml`, `.xhtml`, `.twig`, `.erb`/`.rhtml`, `.aspx`/`.ascx`, `.shtm`/`.shtml`, `.cmp`). The `<?php`/`<%` marker scan stays as a fallback for inline server code in plain `.html`. Both of your repros (`<!-- ${user.name} -->` in `.jsp`, `<!-- <?= $userId ?> -->` in `.php`) are now flagged. Added per-language `comment-only.<ext>` fixtures where the marker lives only inside the HTML comment, so this gap can't get hidden again.

- **ERB syntax**: rspec branch `rule/modify-RSPEC-S1876-narrow-detection` updated to split JSP (`<%-- ... --%>`), ERB (`<%# ... %>`), and Razor (`@* ... *@`); `AvoidHtmlCommentCheck.html` regenerated from that.

Also slipped in: added `#{...}` to the marker set (JSF/Facelets EL was missing), and `.vbhtml` now gets the same Razor handling as `.cshtml`.

vdiez (approved):
Re-reviewed the latest head. The server-side gating gap and the ERB documentation issue are addressed; this now looks good to me.

PR #669, opened 2026-05-27
SONARHTML-243 Add S1097 to Sonar way

erwan-leforestier-sonarsource:
## Summary
This enables HTML rule `InputWithoutLabelCheck` (`RSPEC-1097`) in the built-in Sonar way profile. The source of truth is now the rspec metadata, and the analyzer profile file is the regenerated output from that rspec branch.

## Changes
- Add `Sonar way` to `rules/S1097/html/metadata.json` in rspec
- Regenerate `sonar-html` from the rspec branch so `InputWithoutLabelCheck` is present in `Sonar_way_profile.json`
- Tighten `SonarWayProfileTest` to assert the rule is enabled, and confirm ruling expected issues do not change

## Rule behavior changes in this PR
Before enabling S1097 in Sonar Way we hardened it against common false positives and false negatives:

- `aria-labelledby` now supports multiple space-separated ids and validates each against the document. Binding-form names (`[aria-labelledby]`, `:aria-labelledby`, `v-bind:aria-labelledby`) are validated when the value is a JS string literal (e.g. `"'foo'"`) and trusted when it's an identifier or expression.
- Angular `[id]`/`[for]`/`[attr.id]`, Vue `:id`/`:for`/`v-bind:`, JSX `htmlFor`, and ASP.NET Core `asp-for` are now recognized for label/control association.
- Server-side template expressions in `id`/`for`/`aria-labelledby` values (`${...}`, `{{...}}`, `<%= ... %>`, Razor `@...`, etc.) are trusted and skip the static comparison, in line with how `NoDuplicateIDCheck` already treats dynamic ids.
- **`hasExcludedType` widening**: an `<input>` without an explicit `type` attribute now triggers the rule (per HTML5, missing `type` defaults to `text`, which requires a label). Previously such inputs were skipped. This intentionally adds new violations across the ruling fixtures.

## Functional Validation
Attached: `SONARHTML-243-fv.zip`

Unzip and run:
    ./run.sh

----
## Summary by Gitar

- **Rule logic enhancements:**
  - Improved `InputWithoutLabelCheck` to support dynamic attribute bindings for `id`, `for`, and `aria-labelledby`.
  - Added robust parsing to strip quotes and trim values from property attributes using `getTrimmedPropertyValue`.
  - Enhanced `aria-labelledby` validation to support multiple space-separated IDs and ignore dynamic values.
- **Test coverage:**
  - Added `binding.html` to verify complex attribute binding scenarios and multi-ID `aria-labelledby` validation.
  - Updated `InputWithoutLabelCheckTest` to include new test cases and refined expected violation locations.

### DIFF
--- a/its/ruling/src/test/resources/expected/Web-InputWithoutLabelCheck.json
+++ b/its/ruling/src/test/resources/expected/Web-InputWithoutLabelCheck.json
@@ -86,8 +86,7 @@
 ],
 "project:Silverpeas-Core-master/war-core/src/main/webapp/attachment/jsp/displayAttachedFiles.jsp": [
 1082,
-1083,
-1097
+1083
 ],
 "project:Silverpeas-Core-master/war-core/src/main/webapp/communicationUser/jsp/discussion.jsp": [
 150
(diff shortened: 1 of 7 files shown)


erwan-leforestier-sonarsource:
Functional validation artifact: `SONARHTML-243-fv.zip`

Download:
https://raw.githubusercontent.com/SonarSource/sonar-html/artifacts/SONARHTML-243-fv/SONARHTML-243-fv.zip

erwan-leforestier-sonarsource:
RSPEC source PR: SonarSource/rspec#6927

Regenerated with:
`GITHUB_TOKEN="$(gh auth token)" java -jar /Users/erwan.leforestier/dev/clones/rule-api-2.22.0.5900.jar generate -branch rule/modify-RSPEC-S1097-add-sonar-way -rule InputWithoutLabelCheck`

The regenerated output matches the current PR diff, so no extra analyzer file changes were needed beyond what is already in this branch.

vdiez (changes requested):
I don't think we should enable `InputWithoutLabelCheck` in Sonar way yet.

This PR only enables the rule in `Sonar_way_profile.json`, but it does not harden the analyzer itself. Given the historical noise concern, I think we should first revisit the rule implementation.

A few points that make me uncomfortable enabling it by default:
- `InputWithoutLabelCheck` still reads `for` and `type` through plain `getAttribute(...)`, while the analyzer already has framework-aware property binding support in `TagNode.getProperty(...)`. This means Angular/Vue bindings are not treated consistently here.
- A related label rule (`LabelHasAssociatedControlCheck`) already has explicit support and tests for `[for]`, `:for`, and `v-bind:for`, so the current behavior of S1097 is behind the bar we already apply elsewhere.
- `aria-labelledby` is split with `split(" ")`, which is brittle for multiple spaces or other whitespace.
- The implementation currently excludes `<input>` without a `type`, which does not seem aligned with the rule description.

My preference would be to harden the rule first, add coverage for framework bindings / template cases, and then reconsider enabling it in Sonar way.

vdiez (approved):
The follow-up commits address the concerns from my earlier review before enabling S1097 in Sonar way.

In particular:
- S1097 now uses property-aware handling for id/for/type and covers Angular/Vue binding forms consistently.
- aria-labelledby handling is more robust.
- inputs without an explicit type are now treated as labelable, which matches the rule intent.
- coverage was added for the new binding/template cases.

I’m fine enabling this rule in the default profile now.

PR #670, opened 2026-05-27
SONARHTML-221 Accept lang attribute on <body>

erwan-leforestier-sonarsource:
## Summary
Rule S5254 no longer demands `lang` on `<html>` when `<body>` already carries it. Page-level language scope is now `<html> ∪ <body>`, matching what assistive technologies actually consume.

## Changes
- `LangAttributeCheck` defers the missing-`<html>`-lang verdict until `<body>` is observed: body-lang activates the rule normally, no body (or body without lang) raises the existing violation on `<html>`.
- Updated the violation message to mention both elements (`Add "lang" ... to the "<html>" or "<body>" element`) so users see the new escape hatch.
- Regenerated `S5254.html` from the paired rspec branch `rule/modify-RSPEC-S5254-accept-body-lang`, which adds the missing-lang/body-lang examples to the rule description.
- Extended `LangAttributeCheckTest` fixtures with three new scenarios: html-no-lang + body-lang (compliant), dynamic body-lang (compliant), html + body both missing lang (still non-compliant).

## Functional Validation
Attached: [SONARHTML-221-fv.zip](https://github.com/user-attachments/files/28351614/SONARHTML-221-fv.zip)

Unzip and run:
    ./run.sh


vdiez (changes requested):
Requesting changes for two regressions in `LangAttributeCheck`:

- P1: Deferring activation until a later `<body lang>` skips all pre-body nodes, so invalid `lang` values in `<head>` descendants are no longer reported. Repro: `<html><head><title lang="invalid">Welcome</title></head><body lang="en"><p>Hello</p></body></html>` now yields 0 issues, while the same invalid `<title lang>` under `<html lang="en">...` still yields 1. The updated rule description still says an element with text content and an invalid `lang` value should raise an issue, so this looks like a real regression.

- P2: `hasValidLangAttribute(body)` now reuses the existing broad `hasLangAttribute()` heuristic, and `hasThymeleafLangAttribute()` still treats any `th:attr` containing `lang=` as a language declaration. That creates false negatives such as `<body th:attr="data-lang='en'">`, which now suppresses the missing-page-lang issue even though neither `lang` nor `xml:lang` is actually declared on `<html>` or `<body>`.

I also ran:
- `mvn test -Dtest=LangAttributeCheckTest -pl sonar-html-plugin`
- `mvn test -Dtest=AllowedLangAttributeCheckTest -pl sonar-html-plugin`

vdiez (approved):
Earlier concerns are resolved on the current head.

- The descendant check now runs independently of the deferred page-level `<html>/<body>` decision, so invalid `lang` values in `<head>` descendants are reported again.
- Thymeleaf `th:attr` detection is now anchored to real `lang` / `xml:lang` keys, so `data-lang=` no longer suppresses the page-level issue.

No remaining findings.

Validated with:
- `mvn test -Dtest=LangAttributeCheckTest -pl sonar-html-plugin`
- `mvn test -pl sonar-html-plugin`

I also rechecked the earlier `<head><title lang="invalid">...</title></head><body lang="en">...</body>` regression locally; it now reports the descendant issue as expected.

PR #673, opened 2026-05-27
SONARHTML-255 Track single quotes in tag boundary matcher

erwan-leforestier-sonarsource:
## Summary
Fix S4645 false positive on Vue SFCs where a single-quoted attribute contains `<` (e.g. `:disabled='searchValue.length<1'`). The shared element-boundary matcher only tracked double quotes, so `<` inside `'...'` was counted as a nested tag and the surrounding HTML structure got swallowed.

## Changes
- New `HtmlElementEndMatcher` used only by `NormalElementTokenizer`. The shared `AbstractTokenizer.EndTokenMatcher` is left untouched so JSP/PHP/directive parsing is unaffected.
- The new matcher tracks single-quoted attribute values and treats `<%...%>` as an opaque sub-scope, so `'` in surrounding code (EL string literals, Java/Ruby char literals) doesn't fool the boundary detection.
- New regression tests in `PageLexerTest` and `NestedJavaScriptCheckTest`; ruling integration test is byte-identical to master on the full corpus.

## Functional Validation
Attached: `SONARHTML-255-fv.zip`

Unzip and run:
    ./run.sh

### DIFF
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/lex/HtmlElementEndMatcher.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/lex/HtmlElementEndMatcher.java
@@ -1,0 +1,118 @@
+/*
+ * SonarQube HTML
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * You can redistribute and/or modify this program under the terms of
+ * the Sonar Source-Available License Version 1, as published by SonarSource Sàrl.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+package org.sonar.plugins.html.lex;
+
+import org.sonar.sslr.channel.CodeReader;
+import org.sonar.sslr.channel.EndMatcher;
+
+/*
+ * Detects the end `>` of an HTML element, on top of the basic `<` ... `>` nesting count from
+ * the generic EndTokenMatcher. The generic matcher gets fooled by `<` or `>` characters that
+ * appear inside attribute values or inside JSP/ERB scriptlets; this one tracks three extra
+ * states to avoid that.
+ *
+ * Double-quoted attribute value.
+ *   Opens on `"`, closes on the next `"`. Same as the generic matcher.
+ *
+ * Single-quoted attribute value.
+ *   Opens on `'` only when the previous non-whitespace char is `=`, so a stray apostrophe in
+ *   text or in code (like `Ask'Once` or `l'evenement`) doesn't open a scope. Closes only when
+ *   the next char is an attribute terminator: whitespace, `>`, `/`, `=`, `"`, or EOF. The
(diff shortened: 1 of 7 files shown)


erwan-leforestier-sonarsource:
Functional validation artifact: [SONARHTML-255-fv.zip](https://github.com/SonarSource/sonar-html/raw/refs/heads/artifacts/SONARHTML-255-fv/SONARHTML-255-fv.zip)

vdiez (changes requested):
Blocking regression in `sonar-html-plugin/src/main/java/org/sonar/plugins/html/lex/AbstractTokenizer.java`.

Making `EndTokenMatcher` track single quotes globally changes tokenizers that are not parsing attributes. `CdataTokenizer` now fails on apostrophes inside CDATA. On the current head, `<![CDATA[don't ]]>` no longer closes at `]]>` and runs to EOF.

vdiez (changes requested):
The original regression in AbstractTokenizer is fixed, but the current head still introduces a new parser regression in HtmlElementEndMatcher.

Repro:
<input value='<% if (c == '/') { out.println("x"); } %>' /><p>after</p>

On origin/master, PageLexer returns one closed <input> tag followed by <p>after</p>.
On this PR head, the '/ char literal inside the JSP scriptlet is treated as the end of the outer HTML attribute, so the <input> token is left unterminated and the trailing "' />" is emitted as text.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/lex/HtmlElementEndMatcher.java
```diff
@@ -0,0 +1,114 @@
...
+  private int nesting;
+  private int jspNesting;
+
+  HtmlElementEndMatcher(CodeReader codeReader) {
+    this.codeReader = codeReader;
+  }
+
+  @Override
+  public boolean match(int endFlag) {
+    boolean enteringJsp = previousChar == '<' && endFlag == '%' && activeQuote == 0 && jspNesting == 0;
```
vdiez: enteringJsp only opens the opaque <%...%> scope when activeQuote == 0, so JSP inside a single-quoted attribute never enters JSP mode.

Concrete repro:
<input value='<% if (c == '/') { out.println("x"); } %>' /><p>after</p>

On origin/master this parses as a single closed <input> tag followed by <p>after</p>. On this branch, the '/ char literal closes the outer HTML attribute and leaves the <input> unterminated ("' />" becomes text). The new tests cover top-level JSP and EL inside single-quoted attributes, but not JSP/ERB scriptlets inside quoted attributes.

erwan-leforestier-sonarsource: Good catch — fixed in `b721f35e`. Dropped the `activeQuote == 0` gate from `enteringJsp` so the opaque `<%...%>` scope opens inside quoted attributes too, with the `nesting--` compensation conditional on `activeQuote == 0` so we only undo a nesting bump that actually happened. Added `jsp_scriptlet_inside_single_quoted_attribute_is_opaque` in `PageLexerTest` using your exact repro as the regression guard.

vdiez (approved):
Rechecked the latest head after the follow-up fix.

The JSP-in-single-quoted-attribute regression is addressed now: the matcher enters JSP scope inside quoted attributes, and the exact repro from the previous review is covered by a dedicated lexer test.

I also reran `mvn -pl sonar-html-plugin "-Dtest=PageLexerTest,VueLexerTest,NestedJavaScriptCheckTest" test` on head `3e95e0a851fa6881434ae5ec360b5d6c80279260`.

PR #674, opened 2026-05-27
SONARHTML-246 Splice embedded HTML from PHP strings into the analysis pipeline

erwan-leforestier-sonarsource:
## Summary

HTML inside PHP string literals (`echo "<div role='toolbar'>"`) was invisible to all rules. `PhpEmbeddedHtmlExtractor` now re-lexes string-literal content from PHP directives and splices the resulting nodes (with file-accurate positions) into the main node list before hierarchy construction.

## Gating

Spliced nodes are stamped `embedded=true`. `HtmlAstScanner` routes them only to visitors that implement the new `EmbeddedHtmlCheck` marker interface, keeping context-heavy checks (label association, unclosed tags, doctype…) off the fragmented re-lex output. Six standalone single-element checks opt in: `AriaRoleCheck`, `AriaProptypesCheck`, `ElementWithRoleShouldHaveRequiredPropertiesCheck`, `FocusableInteractiveElementsCheck`, `NoRedundantRolesCheck`, `TabIndexNoPositiveCheck`.

## Changes

- **New** `PhpEmbeddedHtmlExtractor`: handles double-quoted, single-quoted, heredoc and nowdoc literals; replaces interpolations with a dynamic placeholder.
- **New** `EmbeddedHtmlCheck` marker interface + `Node.embedded` flag + one-line gate in `HtmlAstScanner`.
- Six opt-in checks switch from `isDynamicValue` (prefix match) to `containsDynamicValue` (substring match) for mid-value placeholders.
- Ruling baselines updated; non-opt-in baselines reverted to master.

## Functional Validation

See attached [SONARHTML-246-fv.zip](https://github.com/user-attachments/files/28513881/SONARHTML-246-fv.zip). Unzip and run `./run.sh` from inside `SONARHTML-246-fv/`. The script materialises `master` and the PR branch from an embedded bundle, builds each module's classpath, and scans `sample/embedded.php` with `AriaRoleCheck`. Expected delta:

- `master`: no issues — the literal's content never reaches the check.
- PR branch: `Web:S6821` on line 3, the bogus role's actual source line.

### DIFF
--- a/its/ruling/src/test/resources/expected/Web-S6841.json
+++ b/its/ruling/src/test/resources/expected/Web-S6841.json
@@ -1490,8 +1490,11 @@
 11
 ],
 "project:phpMyAdmin-4.0.4-english/libraries/sql_query_form.lib.php": [
+250,
 310,
 313,
-318
+318,
+346,
+365
 ]
 }
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/EmbeddedHtmlCheck.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/EmbeddedHtmlCheck.java
@@ -1,0 +1,44 @@
+/*
+ * SonarQube HTML
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * You can redistribute and/or modify this program under the terms of
+ * the Sonar Source-Available License Version 1, as published by SonarSource Sàrl.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+package org.sonar.plugins.html.checks;
+
+/**
+ * Callback opt-in for checks that want to receive visitor events
+ * ({@code startElement}, {@code endElement}, {@code characters}, …) for HTML
+ * nodes extracted from PHP string literals by {@code PhpEmbeddedHtmlExtractor}.
+ *
+ * <p><b>What this controls:</b> visitor callbacks only. A check that does
+ * <em>not</em> implement this interface will not receive {@code startElement}
+ * etc. for nodes whose {@link org.sonar.plugins.html.node.Node#isEmbedded()}
+ * flag is {@code true}.
+ *
+ * <p><b>What this does not control:</b> AST visibility. The full node list —
+ * including embedded nodes — is still passed to every check's
+ * {@link org.sonar.plugins.html.visitor.DefaultNodeVisitor#startDocument(java.util.List)},
+ * and the parent/child hierarchy remains intact. Non-opted-in checks can
+ * therefore still observe embedded nodes by iterating the {@code startDocument}
+ * list or calling {@link org.sonar.plugins.html.node.Node#getChildren()}, and
+ * can call {@link org.sonar.plugins.html.node.Node#isEmbedded()} to filter when
+ * needed.
+ *
+ * <p>Only standalone single-element or single-attribute checks should opt in.
+ * Checks that reconstruct a DOM from the full node list or that rely on sibling
+ * or ancestor relationships should <em>not</em> opt in; they already reach
+ * embedded nodes through the tree.
+ */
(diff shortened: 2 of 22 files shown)


vdiez (changes requested):
Blocking issue in `sonar-html-plugin/src/main/java/org/sonar/plugins/html/lex/ElementTokenizer.java`.

In the escaped-quoted attribute path, the extra `codeReader.pop()` after the value consumes one character too many and merges adjacent attributes. On the current head, parsing `<x a="x"b="y">` produces a single attribute `a=y` instead of `a=x` and `b=y`.

vdiez (approved):
﻿The original blocker from my May 27 review is resolved.

I re-ran the `<x a="x"b="y">` reproducer on the current head and it now parses as two attributes (`a=x`, `b=y`), so the extra-pop regression in `ElementTokenizer` is no longer present.

One small follow-up remains in `PhpEmbeddedHtmlExtractor.literalTextNode()` (`sonar-html-plugin/src/main/java/org/sonar/plugins/html/lex/PhpEmbeddedHtmlExtractor.java:502`): when a plain text literal sits between two HTML-bearing literals and spans multiple source lines, the generated `TextNode` still gets `endLine = startLine`. For example, `"<a>" . "Hello\nWorld" . "</a>"` currently yields a text node with coordinates `1:20 -> 1:31` instead of spanning onto line 2. This looks low severity and non-blocking, but it should be fixed or covered by a follow-up test.

vdiez (changes requested):
[P2] The new `EmbeddedHtmlCheck` gate only blocks direct callbacks for embedded nodes, but opted-out checks can still observe those nodes through the regular DOM hierarchy.

In `HtmlAstScanner.scan(...)`, embedded nodes are skipped for visitors that do not implement `EmbeddedHtmlCheck`. But `PageLexer.createNodeHierarchy(...)` still attaches those embedded nodes to surrounding real parents, and `TagNode.setParent(...)` eagerly adds them to `parent.getChildren()`. That means an opted-out check can still read embedded HTML indirectly via `getChildren()` / `getParent()`.

I reproduced this with `TableWithoutHeaderCheck`, which is intentionally not allow-listed:

```html
<table>
<?php echo "<th>Header</th>"; ?>
</table>
```

On the current head, `TableWithoutHeaderCheck` reports `0` issues for that file, while the same table without the embedded literal reports `1`. The embedded `<th>` is skipped as a callback target, but it is still present under the real `<table>` node, so `hasHeader()` treats the table as compliant.

If the intent is that opted-out checks ignore PHP-embedded HTML entirely, the boundary needs to apply to hierarchy visibility too, not only visitor callbacks. I think the safer direction is what Victor suggested: embedded nodes should not be automatically exposed as normal children / parents of surrounding real nodes for non-opt-in analysis.

vdiez (approved):
﻿I’m approving this now.

My earlier concern was about the contract, not about the implementation being inherently wrong.

After re-checking the current head, I think the behavior can be defended as follows: `EmbeddedHtmlCheck` is effectively a marker for checks that want to receive embedded nodes through the normal visitor callbacks, while the full AST and node hierarchy remain visible to all checks. That gives rule authors finer control because they can still inspect `Node.isEmbedded()` when traversing from surrounding nodes.

The part that still feels off is that this is not what the current contract says. The `EmbeddedHtmlCheck` javadoc currently says that checks relying on surrounding DOM, sibling relationships, or text content must stay off the embedded path because the re-lexed HTML is fragmented. On the current head, that is no longer strictly true, because non-`EmbeddedHtmlCheck` rules can still observe embedded nodes indirectly via `startDocument(List<Node>)` and `getChildren()`.

According to that current contract, the rules materially affected on this head are:
- `AnchorsHaveContentCheck`
- `ChildElementIllegalCheck`
- `ChildElementRequiredCheck`
- `TableHeaderHasIdOrScopeCheck`
- `TableWithoutCaptionCheck`
- `TableWithoutHeaderCheck`
- `VideoTrackCheck`
- `XPathTemplateCheck`

`XPathTemplateCheck` is the strongest example because it rebuilds a DOM from the full `nodes` list, so the XPath author does not get a chance to filter on `isEmbedded()`.

So I would treat this as a follow-up documentation / contract issue rather than a blocker. Ideally, on top of the current head, I would:

1. Reword `EmbeddedHtmlCheck` and the related scanner comments to make the contract explicit: it is a callback opt-in, not a full embedded-AST visibility opt-in.
2. Add a small regression test that locks this behavior down, so it is clear that non-opted-in checks still see embedded nodes through the tree / `startDocument(...)`.
3. Optionally document the `XPathTemplateCheck` caveat explicitly, since that rule cannot delegate the filtering decision to end users.

With that understanding, the current behavior looks coherent enough to merge.

PR #675, opened 2026-05-27
SONARHTML-384 Skip non-HTML ERB files

erwan-leforestier-sonarsource:
## Summary
Narrow `.erb` file selection in sonar-html so non-HTML ERB templates (`Dockerfile.erb`, `config.yml.erb`, `app.js.erb`, …) are no longer analyzed. The decision is made by a single file predicate that combines two cheap signals — recognized double extension, falling back to a content sniff of the first 2 KB.

## Functional Validation
Attached: [SONARHTML-384-fv.zip](https://github.com/user-attachments/files/28337710/SONARHTML-384-fv.zip)

Unzip and run:
    ./run.sh

----
## Summary by Gitar

- **Refactored file filtering:**
  - Moved ERB file filtering logic from `HtmlSensor` to a new `ErbFileFilter` static predicate.
- **Updated constants:**
  - Added `OTHER_FILE_SUFFIXES` to `HtmlConstants` to include common server-side templates like `php` and `vue`.
- **Enhanced IO utilities:**
  - Added `skipFully` method to `InputFileReader` to robustly advance the reader during content sniffing.

### DIFF
--- a/its/ruling/src/test/resources/expected/Web-BoldAndItalicTagsCheck.json
+++ b/its/ruling/src/test/resources/expected/Web-BoldAndItalicTagsCheck.json
@@ -2951,6 +2951,15 @@
 "project:sonar-master/sonar-server/src/main/webapp/WEB-INF/app/views/users/_autocomplete.html.erb": [
 7
 ],
+"project:sonar-master/sonar-server/src/main/webapp/WEB-INF/gems/gems/actionpack-2.3.15/lib/action_controller/templates/rescues/_request_and_response.erb": [
+17,
+24
+],
+"project:sonar-master/sonar-server/src/main/webapp/WEB-INF/gems/gems/actionpack-2.3.15/lib/action_controller/templates/rescues/template_error.erb": [
+7,
+7,
+11
+],
 "project:sonar-master/sonar-server/src/main/webapp/WEB-INF/gems/gems/rails-2.3.15/lib/rails_generator/generators/components/scaffold/templates/view_show.html.erb": [
 3
 ],
--- a/its/ruling/src/test/resources/expected/Web-BoldAndItalicTagsCheck.json
+++ b/its/ruling/src/test/resources/expected/Web-BoldAndItalicTagsCheck.json
@@ -2951,15 +2951,6 @@
 "project:sonar-master/sonar-server/src/main/webapp/WEB-INF/app/views/users/_autocomplete.html.erb": [
 7
 ],
-"project:sonar-master/sonar-server/src/main/webapp/WEB-INF/gems/gems/actionpack-2.3.15/lib/action_controller/templates/rescues/_request_and_response.erb": [
-17,
-24
-],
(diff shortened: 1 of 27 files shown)


vdiez:
I checked `SONARHTML-384` before replying here. The ticket is intentionally framed as "investigate the best approach", and the only Jira comment also questions whether this filtering should live in `sonar-html` at all or be delegated to the scanner. So I do not think we should lock ourselves too quickly into the current whitelist rule.

Initially the target was to stop analyzing files like `Dockerfile.erb`. This does fix that, but it also skips all bare `.erb` files, which still feels like a filename-only guess. `ERB` is Ruby's general-purpose text templating format, not an HTML-specific one. With no extra context, I would only expect bare `.erb` to generate HTML in a minority of cases, maybe ~25-40%. But that prior changes a lot with context: a bare `.erb` under `views/`, `templates/`, or `app/views/` is much more likely to be HTML. The ruling delta seems to show exactly that kind of file disappearing (`.../app/views/.../*.erb`, `.../templates/rescues/*.erb`), which looks like false negatives rather than just filtering out `Dockerfile.erb`.

So I think the limitation is that we are still purely guessing from the filename, whether that logic sits in `sonar-html` or in the scanner. A stronger and still very cheap approach would be a quick content sniff instead: read only the first few KB, strip ERB blocks (`<% ... %>`), and then apply a very small HTML heuristic on the remaining text.

Concretely, I mean something in this spirit:

```java
String clean = firstChunk.replaceAll("<%[\\s\\S]*?%>", " ");
boolean strongHtml = Pattern.compile("<!DOCTYPE\\s+html\\b|<(?:html|head|body)\\b", Pattern.CASE_INSENSITIVE)
  .matcher(clean)
  .find();
long weakHtmlTags = Pattern.compile("<(?:div|span|p|a|ul|ol|li|table|tr|td|th|form|input|textarea|select|button|script|style|section|article|header|footer|nav|main)\\b", Pattern.CASE_INSENSITIVE)
  .matcher(clean)
  .results()
  .count();

boolean likelyHtml = strongHtml || weakHtmlTags >= 2;
```

The exact regex is not the important part; the point is that this is still computationally cheap: no parser, no full-file read required, just a couple of linear scans on the first `4-8 KB`. It should also be more robust than the current extension whitelist because it would keep obvious HTML ERB templates even when they are named just `foo.erb`, while still rejecting things like `Dockerfile.erb`, `config.yml.erb`, etc.

vdiez (approved):
Reviewed the latest version and I have no blocking findings.

Given the goal here is a best-effort filter to stop analyzing obvious non-HTML `.erb` files such as `Dockerfile.erb`, the current approach looks like a reasonable tradeoff and is clearly better than a hardcoded filename exception. I also ran the focused `ErbFileFilterTest`, `HtmlFileExtensionsTest`, `InputFileReaderTest`, and `HtmlSensorTest` suite locally.

Residual risk remains that some bare `.erb` HTML partials can still be skipped by the heuristic, but that seems acceptable for the scope of this ticket.

PR #678, opened 2026-05-29
SONARHTML-259 Add S7039 Content Security Policy check

erwan-leforestier-sonarsource:
## Summary
Implement S7039 ("Content Security Policies should be restrictive") for the sonar-html plugin. The RSPEC was already defined (csharp implemented this rule) but the HTML side was just a placeholder; this PR adds the Java check, generates the description files from RSPEC, and turns the rule on in the Sonar way profile so it ships out-of-the-box.

## Changes
- New `ContentSecurityPolicyCheck` (`@Rule(key = "S7039")`) scans `<meta http-equiv="Content-Security-Policy">` and the Report-Only variant, tokenises per directive, and flags wildcards, `'unsafe-inline'`, `'unsafe-hashes'`, `'unsafe-eval'` in source-list directives only.
- Directives the browser ignores in meta delivery (`frame-ancestors`, `sandbox`, `report-uri`, `report-to`) are skipped; `'wasm-unsafe-eval'` is exempt.
- Dynamic `content` values are skipped via `Helpers.containsServerSideMarker` plus a CSP-specific Razor detector that tolerates literal `@` in emails and version tokens.
- `Helpers.containsServerSideMarker` extracted as a public helper, shared with `containsDynamicValue`.
- `ContentSecurityPolicyCheckTest` covers all of the above.
- `S7039.html` / `S7039.json` regenerated from RSPEC master via `rule-api`; `Sonar_way_profile.json` picks up the rule automatically.

## Functional Validation
Attached: [SONARHTML-259-fv.zip](https://github.com/user-attachments/files/28398055/SONARHTML-259-fv.zip)

Unzip and run:
    ./run.sh


vdiez (changes requested):
Two false-positive cases need to be addressed before this rule is ready.

- P1: `ContentSecurityPolicyCheck.java:57` reports any `*` anywhere in the `content` attribute, but this rule only analyzes meta-delivered CSPs. In that delivery mode, directives such as `frame-ancestors`, `sandbox`, and `report-uri` are ignored by the browser, so a policy like `default-src 'self'; frame-ancestors *;` is currently flagged even though the wildcard is in a directive that has no effect here.
- P2: `ContentSecurityPolicyCheck.java:60-67` scans the whole serialized policy for `unsafe-inline`, `unsafe-hashes`, and `unsafe-eval` instead of limiting the check to directives where those keywords are actually valid CSP source expressions. I confirmed a local repro with `report-to unsafe-inline`: the rule raises on `unsafe-inline` even though `report-to` takes an endpoint-group token, not a source-list keyword.

I ran `mvn test -Dtest=ContentSecurityPolicyCheckTest -pl sonar-html-plugin` in an isolated worktree. The PR passes as written, and both findings above were confirmed with temporary local repro cases.

erwan-leforestier-sonarsource:
@vdiez addressed in cf49b8f3.

- **P1** (`frame-ancestors *`, `sandbox`, `report-uri`, `report-to`): these directives are now in `META_IGNORED_DIRECTIVES` and skipped entirely before tokens are inspected — the browser ignores them in meta delivery, so we should too. Compliant fixture: `contentSecurityPolicy.html` line 26 (`default-src 'self'; frame-ancestors *; sandbox allow-* allow-scripts; report-uri /reports/*.log; report-to default-endpoint;`).
- **P2** (`report-to unsafe-inline`): the rule now only inspects tokens inside `SOURCE_LIST_DIRECTIVES` (fetch directives + `base-uri`/`form-action`). `report-to`, `trusted-types`, etc. are skipped, and the substring scan over the whole policy is gone. Compliant fixture: `contentSecurityPolicy.html` line 29 (`default-src 'self'; report-to unsafe-inline; trusted-types unsafe-eval policy-a;`).

The implementation now splits on `;`, splits each directive on whitespace, classifies the directive name, then matches tokens as full CSP source expressions (with single-quote stripping). Also fixed in the same commit: `Content-Security-Policy-Report-Only` is now flagged too, and the Razor-skip heuristic was tightened so `mailto:...@...` / `pkg@1.2.3` in `.cshtml` no longer makes the whole CSP skip scanning.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/security/ContentSecurityPolicyCheck.java
```diff
@@ -0,0 +1,149 @@
...
+    String content = node.getAttribute("content");
+    if (content == null || isDynamicCspValue(content)) {
+      return;
+    }
+    reportUnsafeTokens(node, content);
+  }
+
+  private static boolean isCspHeader(String httpEquiv) {
+    return "Content-Security-Policy".equalsIgnoreCase(httpEquiv)
+      || "Content-Security-Policy-Report-Only".equalsIgnoreCase(httpEquiv);
```
vdiez: Content-Security-Policy-Report-Only should not be recognized here. In <meta http-equiv> delivery only Content-Security-Policy is supported, so this would report on tags the browser ignores entirely.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/security/ContentSecurityPolicyCheck.java
```diff
@@ -0,0 +1,149 @@
...
+@Rule(key = "S7039")
+public class ContentSecurityPolicyCheck extends AbstractPageCheck {
+
+  private static final String MESSAGE = "Make sure allowing %s in this Content Security Policy directive is safe here.";
+
+  // A Razor expression starts at an unescaped '@' that begins an identifier, a parenthesised expression,
+  // a code block, or a comment. The lookbehind excludes '@' preceded by an identifier char so that emails
+  // (security@example.com) and version specifiers (pkg@1.0.0) are not mistaken for expressions. Kept local
+  // because Helpers.RAZOR_EXPRESSION is intentionally broader for short-token attributes (lang, role, id).
+  private static final Pattern RAZOR_EXPRESSION_IN_CONTENT = Pattern.compile("(?<![\\w@])@(?!@)[A-Za-z_({*]");
```
vdiez: This still matches static URLs like https://cdn.example.com/@scope/pkg.js in .cshtml. Those are valid static CSP sources, so the rule would skip real violations whenever a path segment starts with @.

vdiez (changes requested):
Two blocker-level correctness issues before this can ship:

- `Content-Security-Policy-Report-Only` is currently treated as a supported `<meta http-equiv>` value. Browsers ignore report-only CSP in meta delivery, so this would raise false positives on tags that have no effect.
- The CSP-specific Razor heuristic still matches static sources like `https://cdn.example.com/@scope/pkg.js` in `.cshtml`. That makes the rule skip real violations whenever a static URL path segment starts with `@`.

I ran `mvn test -pl sonar-html-plugin` on the PR head and it passes, so this looks isolated to the new rule logic.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/security/ContentSecurityPolicyCheck.java
erwan-leforestier-sonarsource: Removed Content-Security-Policy-Report-Only from isCspHeader since browsers ignore that header in meta delivery. Also added a dedicated reportOnlyIsNotFlagged() test with all the unsafe tokens present so the non-regression is explicit.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/security/ContentSecurityPolicyCheck.java
erwan-leforestier-sonarsource: Added / to the negative lookbehind in RAZOR_EXPRESSION_IN_CONTENT, so @scope in a path like cdn.example.com/@scope/pkg.js no longer triggers the Razor skip. New fixture case in contentSecurityPolicy.cshtml covers it.

vdiez (approved):
The two correctness issues from my previous review are fixed.

- `Content-Security-Policy-Report-Only` is no longer treated as a supported `<meta http-equiv>` value.
- The Razor heuristic now keeps static scoped-package URLs like `https://cdn.example.com/@scope/pkg.js` analyzable, and the new tests cover both regressions.

I re-ran `mvn test -pl sonar-html-plugin` on `b595182e`; it passes, and I don’t see remaining review findings on the current head.

PR #680, opened 2026-06-01
SONARHTML-291 Allow S6819 ARIA role whitelist

erwan-leforestier-sonarsource:
## Summary
S6819 can now skip a configured set of ARIA roles when teams intentionally keep those roles instead of switching to the suggested semantic tag.

## Changes
- added a `whitelistedRoles` rule property to `PreferTagOverRoleCheck`
- parse the configured roles once per document and ignore matching roles case-insensitively
- added tests covering the default behavior and configured suppression for `link` and `status`

## Functional Validation
Attached: `SONARHTML-291-fv.zip`

Unzip and run:
    ./run.sh

----
## Summary by Gitar

- **Rule Property:**
  - Added `allowedRoles` rule property to `PreferTagOverRoleCheck` to suppress violations for specific ARIA roles.
- **Role Resolution:**
  - Improved `resolveFirstApplicableRole` to correctly identify the first non-abstract, applicable role from the `role` attribute string.
- **Analysis Warnings:**
  - Implemented `collectAnalysisWarnings` to report invalid role names passed to the `allowedRoles` property.
- **Testing:**
  - Added comprehensive tests for `allowedRoles` configuration, validation of unknown roles, and verifying that only the first applicable role is evaluated.

### DIFF
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/PreferTagOverRoleCheck.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/PreferTagOverRoleCheck.java
@@ -19,12 +19,9 @@ package org.sonar.plugins.html.checks.accessibility;
 import static org.sonar.plugins.html.api.HtmlConstants.hasKnownHTMLTag;
 
 import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.LinkedHashSet;
+import java.util.EnumSet;
 import java.util.List;
 import java.util.Locale;
-import java.util.Set;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


erwan-leforestier-sonarsource:
Functional Validation artifact: [SONARHTML-291-fv.zip](https://github.com/SonarSource/sonar-html/blob/af077e1a88da7ce4343af60b97b1fc702d46bd6e/SONARHTML-291-fv.zip?raw=true)

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/PreferTagOverRoleCheck.java
```diff
@@ -54,4 +93,59 @@ public void startElement(TagNode element) {
...
+  }
+
+  private static Aria.RoleDefinition resolveFirstApplicableRole(String roleAttr) {
+    for (String token : roleAttr.trim().split("\\s+")) {
+      AriaRole role = AriaRole.of(token.toLowerCase(Locale.ROOT));
+      if (role == null) {
+        continue;
+      }
+      Aria.RoleDefinition def = Aria.getRole(role);
+      if (def != null && !def.getElements().isEmpty()) {
```
vdiez: This changes the role-list semantics. `resolveFirstApplicableRole()` currently skips any valid role whose `RoleDefinition` has no mapped HTML elements, but the fallback list should stop at the first valid non-abstract role and only then decide whether S6819 has an equivalent tag to suggest. A repro is `<div role="alert status">`: the explicit role resolves to `alert`, so S6819 should stay quiet, but this code falls through to `status` and reports `<output>`. Please stop at the first valid role here and add a regression test for a concrete role without an element mapping.

vdiez (changes requested):
﻿Blocking issue in fallback-role resolution. See the inline comment for the concrete repro (`role="alert status"`).

vdiez (approved):
﻿Addressed. `resolveFirstApplicableRole()` now stops at the first valid non-abstract role, so cases like `role="alert status"` no longer produce a false positive. I also verified the updated `PreferTagOverRoleCheckTest` passes on the new head.

PR #681, opened 2026-06-01
SONARHTML-329 Support Twig comments

erwan-leforestier-sonarsource:
## Summary
Twig comments are currently lexed as plain text, so markup inside `{# ... #}` can be analyzed as real HTML and trigger false positives such as duplicate IDs. This change recognizes Twig comments in both lexer paths and treats them like server-side comments during scanning.

## Changes
- Tokenize `{# ... #}` comments in both `PageLexer` and `HtmlLexer`.
- Strip Twig comment delimiters in `SonarResolveScanner` so comment directives still parse consistently.
- Add regression coverage for duplicate IDs, AST tokenization, token lexer behavior, and sonar-resolve parsing.

## Functional Validation
Attached: `SONARHTML-329-fv.zip`

Unzip and run:
    ./run.sh

### DIFF
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/lex/PageLexer.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/lex/PageLexer.java
@@ -51,6 +51,8 @@ public class PageLexer {
     new CommentTokenizer("<!--", "-->", true),
     /* JSP Comments */
     new CommentTokenizer("<%--", "--%>", false),
+    /* Twig Comments */
+    new CommentTokenizer("{#", "#}", false),
     /* HTML Directive */
     new DoctypeTokenizer("<!DOCTYPE", ">"),
     /* XML/PHP Directives */
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/whitespace/WhiteSpaceAroundCheck.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/whitespace/WhiteSpaceAroundCheck.java
@@ -65,17 +65,12 @@ public class WhiteSpaceAroundCheck extends AbstractPageCheck {
 
   @Override
   public void comment(CommentNode node) {
-
-    if (node.isHtml()) {
-      if (node.isServerSideInclude()) {
-        return;
-      }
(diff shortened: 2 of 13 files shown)


erwan-leforestier-sonarsource:
Attached: [SONARHTML-329-fv.zip](https://github.com/SonarSource/sonar-html/blob/8800049868932f62d482a0da277a8ca0f7f64541/SONARHTML-329-fv.zip?raw=true)

sonar-html-plugin/src/main/java/org/sonar/plugins/html/lex/PageLexer.java
```diff
@@ -51,6 +51,8 @@ public class PageLexer {
     new CommentTokenizer("<!--", "-->", true),
     /* JSP Comments */
     new CommentTokenizer("<%--", "--%>", false),
+    /* Twig Comments */
+    new CommentTokenizer("{#", "#}", false),
```
vdiez, line 55: Introducing `{# ... #}` as `CommentNode`s makes the existing HTML/JSP split on `CommentNode` incomplete. We still have visitors that treat `!node.isHtml()` as "JSP comment"; `WhiteSpaceAroundCheck` is one example, so Twig comments are now skipped by that rule. I verified that `<%--todo--%>` still raises the 2 expected whitespace issues, while `{#todo#}` raises none. If Twig comments are supposed to behave like the other server-side comments mentioned in the PR description, we probably need explicit delimiter metadata on `CommentNode` or a Twig branch in the affected visitors.

vdiez (changes requested):
I found one regression while validating the new Twig comment handling.

I ran:

`mvn -pl sonar-html-plugin "-Dtest=NoDuplicateIDCheckTest,HtmlLexerTest,PageLexerTest,SonarResolveScannerTest,WhiteSpaceAroundCheckTest,AvoidHtmlCommentCheckTest,NoSonarScannerTest,TodoCommentCheckTest,FixmeCommentCheckTest" test`

vdiez (approved):
Reviewed `9ea38f48`.

The delimiter-based follow-up fixes the previous Twig whitespace regression, and the focused regression suite passed, so this is good to merge.

PR #682, opened 2026-06-01
SONARHTML-371 Add Maven dependency caching for CI

erwan-leforestier-sonarsource:
## Summary
Adds Maven dependency caching and orchestrator caching to the CI workflows so that runners restore cached artifacts instead of re-downloading multiple GBs of data on every build.

The Orchestrator is a SonarSource test library that downloads, starts, and manages a real SonarQube server process for integration tests. On every `qa` and `ruling` run it was downloading the full SonarQube Enterprise distribution ZIP (~110 MB) from Artifactory. The orchestrator cache prevents that re-download by persisting the server directory across runs.

## Changes
- Add `.github/actions/maven-cache/action.yml`: caches `~/.m2/repository` with a monthly+pom-hash key; saves on default branch, restores-only on PR branches
- Add `.github/actions/orchestrator-cache/action.yml`: caches the SonarQube distribution downloaded by the orchestrator; uses a rolling `run_id` key within the month so stale builds are evicted naturally
- Wire both caches into all Maven-invoking jobs in `build.yml`: `build`, `build-windows`, `qa`, `ruling` (orchestrator only for `LATEST_RELEASE`, not `DEV`)
- Wire the maven cache into the `analyze` job in `nightly.yml`

----
## Summary by Gitar

- **CI workflow consolidation:**
  - Integrated nightly schedule into `build.yml` and deleted `nightly.yml`.
  - Added `analyze-shadows` and `iris` jobs for nightly shadow analysis and dogfooding.
- **Test execution updates:**
  - Split `qa` into `qa` and `qa-dev` to separate scheduled SonarQube DEV testing from standard LATEST_RELEASE validation.
  - Updated `qa`, `ruling`, and `promote` jobs to skip execution when triggered by the nightly schedule.
- **Dependency and cache management:**
  - Created `maven-cache` and `orchestrator-cache` actions to reduce download times across `build`, `qa`, `ruling`, and `analyze-shadows`.
  - Automated the removal of stale `sonar-html` artifacts during the Maven cache restoration process.


.github/actions/maven-cache/action.yml
```diff
@@ -0,0 +1,34 @@
...
+  using: composite
+  steps:
+    - name: Compute Maven cache keys
+      id: keys
+      shell: bash
+      run: |
+        HASH=$(find . -name 'pom.xml' -type f | sort | xargs sha256sum | sha256sum | cut -d' ' -f1)
+        MONTH=$(date +'%Y-%m')
+        echo "maven-hash=$HASH" >> $GITHUB_OUTPUT
+        echo "cache-month=$MONTH" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
erwan-leforestier-sonarsource: This is a FP, Windows build job uses Git Bash which ships find, xargs, sha256sum, and rm as Unix-compatible tools under its usr/bin/ directory.

.github/workflows/build.yml
zglicz: as another piece of optimization, could we run DEV qa only on master during the nightly? This is sth we already do for SonarJS, as when we have a high load of PRs, this wouldn't hit any cache.

zglicz (approved):
Nice job. 1 small improvement. Feel free to update it here or in a follow-up PR.

.github/workflows/build.yml
erwan-leforestier-sonarsource: Done, I added an `if` condition on the `qa` job that skips the `DEV` matrix entry on non-default branches. Since `build.yml` has no schedule trigger (unlike SonarJS), I gated it on `github.ref_name == github.event.repository.default_branch` which has the same effect: DEV only runs after a merge to master.Done, I added an `if` condition on the `qa` job that skips the `DEV` matrix entry on non-default branches. Since `build.yml` has no schedule trigger (unlike SonarJS), I gated it on `github.ref_name == github.event.repository.default_branch` which has the same effect: DEV only runs after a merge to master.

zglicz: Hm... that still makes it so that it runs on each merge to master. I'd suggest it to be a nightly thing, because it is not really also a valuable piece of information on a merge to master. So say we have not much development on this plugin, we might not even notice that the current DEV of SQ is not working with our analyzer. And on the contrary, if we have a lot of development, we would keep on trying to download a new version of SQ based on their master.

See how SonarJS has it setup. sonar-html also has a nightly build, which currrently is for iris, but they could be unified, similar to SonarJS perhaps.

erwan-leforestier-sonarsource: Ok I see, I've pushed one commit [f59ea0041](https://github.com/SonarSource/sonar-html/pull/682/changes/f59ea00416530ed702026fd4cac95c1a2ff61029) to combine  nightly.yml with build.yml and run the DEV QA only on master on schedule to run nightly.

.github/workflows/build.yml
```diff
@@ -49,12 +54,9 @@ jobs:
           maven-args: '-T1C'
 
   qa:
-    if: github.actor != 'dependabot[bot]'
+    if: github.actor != 'dependabot[bot]' && github.event_name != 'schedule'
```
zglicz: I think it's fine if it runs on the scheduled run. Lets make the nightly all-encompassing.

.github/workflows/build.yml
```diff
@@ -74,17 +76,38 @@ jobs:
...
+      - *maven_cache
+      # No orchestrator cache: SQ:DEV moves daily, the cache key would never hit.
+      - name: Run QA tests against SQ:DEV
+        env: *qa_env
+        run: |
+          mvn -f its/pom.xml -pl plugin,sonarlint-tests verify -Pqa -Dsonar.runtimeVersion=DEV -Dmaven.test.redirectTestOutputToFile=false -B -e -V
 
   ruling:
-    if: github.actor != 'dependabot[bot]'
+    if: github.actor != 'dependabot[bot]' && github.event_name != 'schedule'
```
zglicz: same

.github/workflows/build.yml
```diff
@@ -196,7 +221,52 @@ jobs:
...
+        with:
+          primary_project_key: org.sonarsource.html:html
+          primary_platform: Next
+          shadow1_project_key: SonarSource_sonar-html
+          shadow1_platform: SQC-EU
+          shadow2_project_key: SonarSource_sonar-html
+          shadow2_platform: SQC-US
+
   promote:
+    if: github.event_name != 'schedule'
```
zglicz: lets also remove

zglicz (approved):
few places to remove the scheduled check

.github/workflows/build.yml
erwan-leforestier-sonarsource: Done, removed the guard on qa, ruling, and promote — the nightly now runs everything.

.github/workflows/build.yml
erwan-leforestier-sonarsource: Done.

.github/workflows/build.yml
erwan-leforestier-sonarsource: Done.

zglicz:
Looks good!

PR #683, opened 2026-06-01
SONARHTML-289 Add rule S8720 for aria-role attribute usage

erwan-leforestier-sonarsource:
## Summary
Adds new HTML rule S8720 that flags use of the non-standard `aria-role` attribute. Browsers and assistive technologies do not recognize it, so the intended role is silently ignored — `role` is the correct WAI-ARIA attribute name.
RSPEC PR: https://github.com/SonarSource/rspec/pull/6955

## Changes
- New `AriaRoleAttributeCheck` raising an issue whenever an element carries an `aria-role` attribute (case-insensitive name match, also catches Angular/Vue property-binding variants such as `[aria-role]`, `:aria-role`).
- Generated rule description files (`S8720.html`/`S8720.json`) via `rule-api`; enabled in the Sonar way profile.
- Unit tests covering noncompliant cases (different elements, case variants, empty value, value-less attribute), compliant cases (`role`, unrelated `aria-` attributes), and framework binding variants.

## Functional Validation
Attached: [SONARHTML-289-fv.zip](https://github.com/user-attachments/files/28468678/SONARHTML-289-fv.zip)

Unzip and run:
    ./run.sh

### DIFF
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/AriaRoleAttributeCheck.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/AriaRoleAttributeCheck.java
@@ -1,0 +1,35 @@
+/*
+ * SonarQube HTML
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * You can redistribute and/or modify this program under the terms of
+ * the Sonar Source-Available License Version 1, as published by SonarSource Sàrl.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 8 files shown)


vdiez (approved):
﻿Reviewed together with the associated RSPEC update in SonarSource/rspec#6955.

The implementation, generated metadata, and focused tests look good to me.

Non-blocking follow-up: S8720 currently covers literal HTML plus the Angular/Vue binding aliases already normalized by `TagNode`, but it does not yet catch Thymeleaf-generated variants such as `th:aria-role` or `th:attr="aria-role=..."`. If we want that coverage, I'd handle it in a follow-up and widen RSPEC in lockstep.

PR #686, opened 2026-06-03
SONARHTML-386 USER-2103 Fix S6840 false positives on input without type and numeric control groups

erwan-leforestier-sonarsource:
## Summary
Rule S6840 raised false positives on `<input>` elements without an explicit `type` attribute (including those using Razor's `asp-for` tag helper), and on `<input type="number">` for five autocomplete values that were incorrectly mapped to the Multiline control group instead of Numeric.

## Changes
- `ControlGroup.belongsToTextControlGroup`: treat `<input>` with no `type` as Text state (HTML spec default), so it belongs to the text control group
- `ValidAutocompleteCheck`: correct control group for `cc-exp-year`, `transaction-amount`, `bday-day`, `bday-month`, `bday-year` from Multiline to Numeric per WHATWG HTML spec
- Update existing test files to remove false-noncompliant annotations on type-less inputs; split parameterised test to reflect that `street-address` (Multiline) behaves differently; add five new test files for the corrected field names

## Functional Validation
Attached: [USER-2103-fv.zip](https://github.com/user-attachments/files/28549698/USER-2103-fv.zip)

Unzip and run:
    ./run.sh

### DIFF
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/ControlGroup.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/ControlGroup.java
@@ -16,15 +16,27 @@
  */
 package org.sonar.plugins.html.api.accessibility;
 
+import java.util.Locale;
+import java.util.Set;
+import javax.annotation.CheckForNull;
 import org.sonar.plugins.html.node.TagNode;
 
 public class ControlGroup {
+  private static final String INPUT_TAG = "input";
+  // All valid HTML input type values per WHATWG; anything not in this set has an
+  // invalid-value default of the Text state.
+  private static final Set<String> VALID_INPUT_TYPES = Set.of(
+    "hidden", "text", "search", "tel", "url", "email", "password",
+    "date", "month", "week", "time", "datetime-local", "number", "range", "color",
+    "checkbox", "radio", "file", "submit", "image", "reset", "button"
+  );
+
   public static boolean belongsToAutofillExpectationMantleControlGroup(TagNode node) {
-    if (!node.getNodeName().equalsIgnoreCase("input")) {
+    if (!node.getNodeName().equalsIgnoreCase(INPUT_TAG)) {
       return true;
     }
 
-    var type = node.getAttribute("type");
+    var type = resolveType(node);
 
     return type == null || !type.equalsIgnoreCase("hidden");
   }
@@ -34,13 +46,13 @@ public class ControlGroup {
       return true;
     }
 
-    var type = node.getAttribute("type");
+    var type = resolveType(node);
 
     return type != null && type.equalsIgnoreCase("date");
   }
 
   public static boolean belongsToMonthControlGroup(TagNode node) {
-    var type = node.getAttribute("type");
+    var type = resolveType(node);
 
     if (type != null && type.equalsIgnoreCase("month")) {
       return true;
@@ -56,17 +68,17 @@ public class ControlGroup {
       return true;
     }
 
-    if (!nodeName.equalsIgnoreCase("input")) {
+    if (!nodeName.equalsIgnoreCase(INPUT_TAG)) {
       return false;
     }
 
-    var type = node.getAttribute("type");
+    var type = resolveType(node);
 
     return type != null && type.equalsIgnoreCase("hidden");
   }
 
   public static boolean belongsToNumericControlGroup(TagNode node) {
-    var type = node.getAttribute("type");
+    var type = resolveType(node);
 
     if (type != null && type.equalsIgnoreCase("number")) {
       return true;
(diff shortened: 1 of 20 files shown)


sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/ControlGroup.java
```diff
@@ -102,10 +113,15 @@ public static boolean belongsToTextControlGroup(TagNode node) {
...
+    if (!nodeName.equalsIgnoreCase(INPUT_TAG)) {
+      return false;
+    }
+
     var type = node.getAttribute("type");
 
-    if (type == null) {
-      return false;
+    // Both missing and invalid type values default to the Text state per WHATWG.
+    if (type == null || !VALID_INPUT_TYPES.contains(type.toLowerCase(Locale.ROOT))) {
```
vdiez: [P1] This branch now treats any unresolved `type` as a Text control, but it still reads the value with `getAttribute("type")`. In this parser, Angular/Vue bindings like `[type]`, `:type`, and `v-bind:type` are only exposed through `getProperty(...)` / `getPropertyValue(...)`, so inputs such as `<input [type]="'checkbox'" autocomplete="organization">` are seen as `type == null` here and incorrectly accepted. I verified this with a focused test: the rule no longer raises an issue for statically-known bound non-text types. We should either read and normalize bound literal `type` values here, or avoid treating unresolved dynamic bindings as Text controls.

vdiez (changes requested):
Requesting changes because the new `type` handling introduces a regression in S6840 for statically-known bound `type` values.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/ControlGroup.java
erwan-leforestier-sonarsource: Thanks for catching that! You were right, `getAttribute("type")` was missing Angular/Vue bound attributes. The fix uses `getProperty("type")` instead, which covers all binding forms (`[type]`, `:type`, `v-bind:type`). Bound string literals like `[type]="'checkbox'"` are now unwrapped and checked against the control group. For dynamic expressions that cannot be statically resolved, the rule is skipped entirely on that element.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/ValidAutocompleteCheck.java
```diff
@@ -327,6 +327,10 @@ public void startElement(TagNode node) {
       return;
     }
 
+    if (ControlGroup.hasDynamicTypeBinding(node)) {
+      return;
+    }
```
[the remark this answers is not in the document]
erwan-leforestier-sonarsource: on purpose to not raise FP

vdiez (approved):
Re-reviewed the latest changes. The bound-literal handling fixes my original blocking concern, and I agree with skipping S6840 on inputs whose type is dynamically bound to avoid false positives. No blocking issues remain from my side.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/ValidAutocompleteCheck.java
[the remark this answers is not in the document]
vdiez: Non-blocking edge case: `hasDynamicTypeBinding()` / `isStaticStringLiteral()` check the raw bound value without trimming first. That means a statically-known literal like `<input [type]=" 'checkbox' " autocomplete="organization">` is classified as dynamic instead of resolvable, and `startElement()` then skips S6840 entirely for that node. Since the intent here seems to be "unwrap bound string literals, skip only genuinely unresolved expressions", it may be worth trimming before the literal check so whitespace-padded literals are still analyzed.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/ControlGroup.java
```diff
@@ -131,6 +154,50 @@ public static boolean belongsToUsernameControlGroup(TagNode node) {
...
 
+  // Returns true when the node has a bound type attribute (e.g. [type], :type) whose
+  // value is a dynamic expression that cannot be statically resolved. String literals
+  // like [type]="'checkbox'" are resolvable and therefore not dynamic.
+  public static boolean hasDynamicTypeBinding(TagNode node) {
+    var typeAttr = node.getProperty("type");
+    if (typeAttr == null || typeAttr.getName().equalsIgnoreCase("type")) {
+      return false;
+    }
+    return !isStaticStringLiteral(typeAttr.getValue());
```
vdiez, line 165: Non-blocking edge case: this literal-vs-dynamic check runs on the raw bound value without trimming first. A statically-known literal like `<input [type]=" 'checkbox' " autocomplete="organization">` therefore fails `isStaticStringLiteral(...)`, gets classified as dynamic, and S6840 is skipped for that node. Since the intent here seems to be "unwrap bound string literals, skip only genuinely unresolved expressions", trimming before this check would let whitespace-padded literals stay analyzable.

PR #689, opened 2026-06-03
SONARHTML-245 Narrow S1101 to sibling same-text links only

erwan-leforestier-sonarsource:
## Summary
Restricts rule S1101 to flag only same-text different-target links that share the same direct parent element (siblings). Cross-container links with identical text are now compliant because surrounding context provides disambiguation per WCAG 2.4.4. Also adds the rule to the Sonar way profile.
RSPEC PR: https://github.com/SonarSource/rspec/pull/7000

## Changes
- Switch link tracking from a document-wide map to a per-parent `IdentityHashMap<TagNode, Map<String, Link>>`
- Capture the parent element (`node.getParent()`) when entering each `<a>` tag
- Add test cases for different-parent (compliant) and same-parent (non-compliant) scenarios
- Add `LinksIdenticalTextsDifferentTargetsCheck` to `Sonar_way_profile.json`
- Regenerate rule description from rspec branch

## Functional Validation
Attached: `SONARHTML-245-fv.zip`

Unzip and run:
    ./run.sh


vdiez (changes requested):
I rechecked the latest head (`4a007420`), including the follow-up ruling update and the Sonar Way profile change. The earlier findings still reproduce on the current tip.

1. In `sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/LinksIdenticalTextsDifferentTargetsCheck.java` (around lines 55 and 78), bucketing links only by direct parent identity is too narrow and misses real S1101 cases behind neutral wrappers. On the current head, `<div><span><a href="/user/1/edit">Edit</a></span><span><a href="/user/2/edit">Edit</a></span></div>` produces `0` issues, even though the links are still indistinguishable.

2. The implementation still flags links whose purpose is disambiguated by surrounding text, which does not match the updated rule documentation. On the current head, `<div>Alice: <a href="/alice">Edit</a> Bob: <a href="/bob">Edit</a></div>` still raises `1` issue, even though the surrounding text makes each link's purpose clear.

3. `sonar-html-plugin/src/test/resources/checks/LinksIdenticalTextsDifferentTargetsCheck.html` now contains standalone `<li>` elements, which trips `ItemTagNotWithinContainerTagCheck` and adds avoidable analyzer noise to the fixture.

I also reran `mvn test -Dtest=LinksIdenticalTextsDifferentTargetsCheckTest -pl sonar-html-plugin` on the latest head; it passes.

Because the latest commit also adds this rule to `Sonar_way_profile.json`, I think we should resolve the rule semantics before enabling it by default.

erwan-leforestier-sonarsource:
@vdiez

> 1. In `sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/LinksIdenticalTextsDifferentTargetsCheck.java` (around lines 55 and 78), bucketing links only by direct parent identity is too narrow and misses real S1101 cases behind neutral wrappers. On the current head, `<div><span><a href="/user/1/edit">Edit</a></span><span><a href="/user/2/edit">Edit</a></span></div>` produces `0` issues, even though the links are still indistinguishable.


As described on ticket SONARHTML-245, we agreed on narrowing down the rule to compare only links under the same parent element to prevent raising FP. Rationale is the following: capturing such ambiguous same-text links would require to compare all the links to others as it was before, leading to some FP raised when 2 links dispatched on different sections of the same page share the same text but have different links.

erwan-leforestier-sonarsource:
@vdiez

> The implementation still flags links whose purpose is disambiguated by surrounding text, which does not match the updated rule documentation. On the current head, `<div>Alice: <a href="/alice">Edit</a> Bob: <a href="/bob">Edit</a></div>` still raises `1` issue, even though the surrounding text makes each link's purpose clear.

There is no way to know if text around disambiguates the context or not. I can take the same example but with different text around like  `<div>Congratulation to <a href="/alice">this user</a> who collaborated with <a href="/bob">this user</a></div>`. As agreed on ticket SONARHTML-245, this is considered as edge case and we want a rule violation to be raised here so the developer can demystify if it's indeed a FP or not.

vdiez (approved):
I rechecked the latest head and the follow-up discussion.

The concrete issues I had raised are now addressed or clarified:
- the fixture noise from orphaned `<li>` elements is fixed
- the consecutive-conflict case is now covered correctly
- the remaining direct-parent / surrounding-text behavior is intentional and aligns with the agreed scope of SONARHTML-245

Approving based on the current implementation and discussion.

PR #690, opened 2026-06-04
SONARHTML-264 Add rule S8732: legend must be direct child of fieldset

erwan-leforestier-sonarsource:
## Summary
Add rule S8732 to flag `<legend>` elements that are not a direct child of `<fieldset>`. Screen readers cannot associate a legend with its fieldset group when the legend is wrapped in an intermediate `<div>` or other element.
RSPEC PR: https://github.com/SonarSource/rspec/pull/7005

## Changes
- Add `LegendMustBeDirectChildOfFieldsetCheck` (rule S8732): reports any `<legend>` whose direct parent is not `<fieldset>`
- Add generated rule resources `S8732.html` / `S8732.json` from rspec branch
- Update `Sonar_way_profile.json` to include S8732
- Add test fixtures `test.html` and `test.vue` covering compliant and noncompliant cases

## Functional Validation
Artifact: [Uploading SONARHTML-264-fv.zip…]()

    ./run.sh

----
## Summary by Gitar

- **New Feature:**
  - Added `MisplacedLegendCheck` (S8732) to enforce `<legend>` is a direct child of `<fieldset>` or `<optgroup>`.
  - Added `RazorSectionScopeTracker` to prevent false positives by tracking Razor `@section` blocks.
- **Refactored Logic:**
  - Integrated `RazorSectionScopeTracker` into `ItemTagNotWithinContainerTagCheck` to suppress warnings in scoped fragments.
  - Enhanced `Helpers.isTemplateLikeTag` to identify `asp:*` and `ng-template` as transparent wrappers.
- **Documentation & Tests:**
  - Added rule metadata, `Sonar_way_profile.json` updates, and comprehensive `html`/`vue`/`cshtml` test fixtures.

### DIFF
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/Helpers.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/Helpers.java
@@ -138,7 +138,15 @@ public class Helpers {
     return hasAncestorMatching(node, Helpers::isTemplateLikeTag);
   }
 
-  private static boolean isTemplateLikeTag(TagNode node) {
+  /**
+   * Returns true if {@code node} is a template-like wrapper: HTML {@code <template>},
+   * Angular {@code <ng-template>}, or an ASP.NET WebForms server control ({@code asp:*}).
+   * These elements do not contribute to the rendered DOM and can be treated as transparent.
+   *
+   * @param node the tag node to test
+   * @return true if the node is a template-like wrapper
+   */
+  public static boolean isTemplateLikeTag(TagNode node) {
     String name = node.getNodeName();
     if (name == null || name.isEmpty()) {
       return false;
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/RazorSectionScopeTracker.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/RazorSectionScopeTracker.java
@@ -1,0 +1,129 @@
+/*
+ * SonarQube HTML
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * You can redistribute and/or modify this program under the terms of
+ * the Sonar Source-Available License Version 1, as published by SonarSource Sàrl.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+package org.sonar.plugins.html.api;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.sonar.plugins.html.node.Node;
+import org.sonar.plugins.html.node.TagNode;
+import org.sonar.plugins.html.node.TextNode;
+import org.sonar.plugins.html.visitor.HtmlSourceCode;
+
+/**
+ * Tracks top-level Razor {@code @section ... { ... }} scopes so checks can suppress
+ * false positives on markup whose rendered container lives outside the current file fragment.
+ */
+public final class RazorSectionScopeTracker {
+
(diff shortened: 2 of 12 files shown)


vdiez (commented):
﻿I found a few issues before we add `S8732`:

- The implementation flags every `<legend>` whose direct parent is not `<fieldset>`, but `<legend>` is also allowed as the first child of `<optgroup>` in customizable select markup. As written, this will raise false positives on valid markup.
- Similar structural checks already exempt template-like scopes (`<template>`, `<ng-template>`, `asp:*`) and Razor sections. This rule does not, so it is likely to report false positives inside inert/template content.
- The rule description says the `<legend>` should be the first child of `<fieldset>`, but the implementation only enforces direct parentage. Either the implementation needs an ordering check or the wording needs to be narrowed.

Non-blocking note: there is already an exact generic template for this pattern in `ParentElementRequiredCheck` (`child=legend`, `parent=fieldset`), plus `ItemTagNotWithinContainerTagCheck` covers the same family of structural constraints. I am fine with a dedicated rule if we want it in Sonar way, but I would not present it as a brand new pattern.

vdiez:
One more suggestion for `S8732`: I would avoid raising when the `<legend>` is under an unknown/custom parent element.

This repo already uses `hasKnownHTMLTag(...)` in several accessibility rules to avoid false positives on framework or postprocessed tags. For this rule, that would give a safer behavior:

- parent is `fieldset` or `optgroup`: compliant
- parent is a known HTML tag other than those: raise
- parent is an unknown/custom tag: skip
- template-like scopes / Razor sections: skip, same idea as `ItemTagNotWithinContainerTagCheck`
- no parent: still raise

I think that is a better fit for template-heavy files than treating every non-`fieldset` parent as definitely wrong.

vdiez (commented):
I still see two issues in the latest update:

- `sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/LegendMustBeDirectChildOfFieldsetCheck.java:39-50`: the rule now treats `<template>` / `<ng-template>` as transparent, but it still raises on unknown/custom wrapper parents. Existing accessibility rules usually skip non-native tags via `hasKnownHTMLTag(...)`, so this can still produce false positives for postprocessed wrappers such as `<fieldset><my-wrapper><legend>...</legend></my-wrapper></fieldset>` or `asp:*` controls.
- `sonar-html-plugin/src/main/resources/org/sonar/l10n/web/rules/Web/S8732.json:2`: the metadata title still says only `"<fieldset>" elements`, while the implementation and `S8732.html` now allow both `<fieldset>` and `<optgroup>`.

erwan-leforestier-sonarsource:
@vdiez Both addressed in the latest commits, thanks for the second pass.

1. The `hasKnownHTMLTag(...)` skip is already in place at line 54 since `feccf582`. Your `<fieldset><my-wrapper><legend>...</legend></my-wrapper></fieldset>` example is compliant under the current code: `my-wrapper` is not in `KNOWN_HTML_TAGS` so we skip. Added a fixture matching that exact shape in `bd8a632c` so it stays covered.

2. Good catch on the title. Fixed in rspec and regenerated; `S8732.json` now reads `"<legend>" tags should be direct children of "<fieldset>" or "<optgroup>" elements`.

vdiez (changes requested):
One remaining blocker before I’d approve:

- [P2] `sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/MisplacedLegendCheck.java:37-46`
  The rule still raises on a parentless `<legend>` without any Razor-specific escape hatch. That leaves a confirmed false positive on supported `.cshtml` input. I reproduced it on the PR head with:

  `@section LegendSection {`
  `  <legend>Group title</legend>`
  `}`

  This is reported today even though the enclosing `<fieldset>` may only exist at render time, which matches the class of framework/template false positives the rule is already trying to avoid for `<template>`, `asp:*`, and unknown tags.

Please either skip Razor fragment/section scopes here and add a regression test, or document that `.cshtml` fragment scopes are intentionally unsupported and keep the rule out of Sonar way until that limitation is acceptable.

vdiez (approved):
The latest commit addresses my blocker.

- `MisplacedLegendCheck` now skips parentless `<legend>` elements that sit inside Razor `@section ... { ... }` scopes.
- The new `test.cshtml` regression covers the reported false-positive shape.
- I also re-ran `MisplacedLegendCheckTest` and `ItemTagNotWithinContainerTagCheckTest` on the updated head to verify the shared `RazorSectionScopeTracker` extraction did not regress the existing Razor-section handling.

PR #691, opened 2026-06-04
SONARHTML-178 Skip out-of-range issues instead of crashing

erwan-leforestier-sonarsource:
## Summary
When the HTML lexer assigns an issue a line number beyond the file's actual line count, locationForIssue() now skips the issue with a WARN log instead of throwing IllegalArgumentException. Previously the exception propagated through saveMetrics() and was caught by the outer execute() try-catch, which dropped every issue and metric from the affected file and produced an AnalysisError in the SQ UI.

## Changes
- Add bounds-checking in locationForIssue() that returns null for out-of-range lines instead of crashing
- Add isValidLine() helper for the bounds check
- Fix unreadable_file test which accidentally relied on the crash to produce its AnalysisError

## Functional Validation
Artifact: SONARHTML-178-fv.zip

    ./run.sh

Expected output is in expected-output.txt. The README shows the before/after comparison so you can reproduce the difference directly.

:warning::warning: This is not ready for review :warning::warning:

### DIFF
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/core/HtmlSensor.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/core/HtmlSensor.java
@@ -136,7 +136,7 @@ public final class HtmlSensor implements Sensor {
     }
   }
 
-  private static void saveMetrics(SensorContext context, HtmlSourceCode sourceCode) {
+  static void saveMetrics(SensorContext context, HtmlSourceCode sourceCode) {
     InputFile inputFile = sourceCode.inputFile();
 
     for (Map.Entry<Metric<Integer>, Integer> entry : sourceCode.getMeasures().entrySet()) {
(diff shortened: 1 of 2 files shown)


erwan-leforestier-sonarsource:
Closing this PR.

After investigating SONARHTML-178 in depth, the underlying bug does not reproduce on current master:

- Ran the full HtmlSensor with every rule active against the exact file from the ticket (moodle `lib/filterlib.php` at commit `e77452220b`). Result: 7 issues raised, **0 analysis errors**. No `IllegalArgumentException`.
- The lexer's max line position for any node on that file is 1715, which is in range for `inputFile.lines() = 1715`.

The original symptom (issue at line 2320 on a 1715-line file — a 605-line drift) is consistent with lexer state corruption from a PHP directive not closing properly. SONARHTML-169 (#591) added `CodeAwareEndMatcher` to `AbstractTokenizer` to handle `?>` correctly inside string literals and comments, which is the most likely fix — though I have not bisected to confirm it specifically.

The approach in this PR — bounds-checking in `HtmlSensor.locationForIssue` and dropping out-of-range issues with a WARN — has two problems:

1. It does not fix any currently-failing scenario; the synthetic tests inject `HtmlIssue(ruleKey, 9999, ...)` directly rather than exercising any real lexer code path.
2. It would mask future lexer regressions: an issue silently disappearing into a WARN log is much harder to notice than a hard failure.

The right follow-up, if we want a regression guard, is a sensor-level test that lexes the actual `filterlib.php` and asserts `allAnalysisErrors().isEmpty()`. That locks in the existing fix without burying future bugs.

PR #693, opened 2026-06-05
SONARHTML-388 USER-2104 Handle empty Razor asp-for labels

erwan-leforestier-sonarsource:
## Summary
Handle Razor `label[asp-for]` in `S6853` the same way ASP.NET Core does: empty or whitespace-only label bodies get fallback text at render time, while non-empty bodies are still analyzed normally.

## Changes
- treat empty or whitespace-only `label[asp-for]` bodies as having generated accessible text
- keep reporting `label[asp-for]` when nested content suppresses the generated label text
- extend the `S6853` tests with empty, whitespace-only, nested tag, and comment-only Razor label cases

## Functional Validation
Artifact: [USER-2104-fv.zip](https://github.com/user-attachments/files/28637675/USER-2104-fv.zip)

    ./run.sh


vdiez (approved):
Approving.

Checked against USER-2104 and the analyzer policy for dynamic/server-generated markup. In this Razor `label[asp-for]` case, the accessible text is generated outside the static artifact, so avoiding `S6853` findings for empty / whitespace-only bodies is the right behavior.

PR #715, opened 2026-06-24
SONARHTML-395 Handle template-generated labels in S6853

erwan-leforestier-sonarsource:
## Summary
Handle `S6853` false positives on template-generated labels by recognizing Razor and Thymeleaf patterns that only resolve at render time.

## Changes
- track whether a Razor `asp-for` label is truly empty before trusting its generated text
- recognize Thymeleaf `th:for` and `th:attr` associations, plus template text attributes such as `th:text`, `th:utext`, `v-text`, and `v-html`
- add regression coverage for empty Razor labels and template-generated label content

## Functional Validation
Artifact: [SONARHTML-395-fv.zip](https://github.com/user-attachments/files/29328351/SONARHTML-395-fv.zip)

    ./run.sh

Expected output:
```text
Analyzing "razor-empty-label.cshtml"...
Results:
    - Rule "S6853" -> L.1
Analyzing "template-generated-label.html"...
Results:
    - Rule "S6853" -> L.1
    - Rule "S6853" -> L.2

****** Branch "fix/sonarhtml-395-handle-template-generated-labels" ******
Analyzing "razor-empty-label.cshtml"...
Results:
    (none)
Analyzing "template-generated-label.html"...
Results:
    (none)
```

----
## Summary by Gitar

- **Improved template handling:**
  - Integrated `AccessibilityUtils` and `Thymeleaf` helpers to identify dynamic content in `HeadingHasAccessibleContentCheck` and `LabelHasAssociatedControlCheck`.
  - Added support for Angular binding properties `[innerText]`, `[innerHTML]`, and `[textContent]` as valid accessible label text.
  - Updated `LabelHasAssociatedControlCheck` to treat HTML comments as transparent rather than blocking label content.
- **Refined association logic:**
  - Updated `hasPropertyHint` to correctly identify template-generated associations even when values are empty.
  - Restricted Razor fragment detection logic to Razor-specific files to avoid false positives in standard HTML.
  - Explicitly ignore `for=""` or whitespace-only associations, which are no longer considered valid controls.
- **Test coverage:**
  - Added extensive regression tests for template-generated labels, Razor-specific fragment rendering, and Angular binding patterns.

### DIFF
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/Thymeleaf.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/Thymeleaf.java
@@ -18,6 +18,10 @@ package org.sonar.plugins.html.api;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
+import java.util.Set;
+import javax.annotation.Nullable;
+import org.sonar.plugins.html.node.Attribute;
 import org.sonar.plugins.html.node.TagNode;
 
 /**
@@ -32,7 +36,27 @@ import org.sonar.plugins.html.node.TagNode;
  */
 public final class Thymeleaf {
 
+  /**
+   * Attributes that ask Thymeleaf to render an external fragment in place of (or inside) the
+   * current element. Statically, the resulting markup is opaque — the rendered fragment can
+   * supply text, controls, or both — so checks should treat these as "do not flag".
+   */
+  public static final Set<String> FRAGMENT_INSERTION_ATTRIBUTES = Set.of("th:insert", "th:include", "th:replace");
+
   private Thymeleaf() {
+  }
+
+  /**
+   * Returns whether the node carries any Thymeleaf fragment-insertion attribute
+   * ({@code th:insert}, {@code th:include}, {@code th:replace}).
+   */
+  public static boolean hasFragmentInsertion(TagNode node) {
+    for (Attribute attribute : node.getAttributes()) {
+      if (FRAGMENT_INSERTION_ATTRIBUTES.contains(attribute.getName().toLowerCase(Locale.ROOT))) {
+        return true;
+      }
+    }
+    return false;
   }
 
   /**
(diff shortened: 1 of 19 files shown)


asya-vorobeva (commented):
It looks like corresponding RSpec description is very confusing: https://musical-adventure-r9qk65j.pages.github.io/rspec/#/rspec/S6853/html
It says that label elements should have a text label (which is obviously wrong for template-generated labels) and doesn't provide list of exceptions. Should we update it?

erwan-leforestier-sonarsource:
Hi @asya-vorobeva thanks for the review, it's true that the rspec rule description is bad, but it was improved in this PR: https://github.com/SonarSource/rspec/pull/7179

Though there is no exception listed in the rule description because I think that either you fall into one of the exception, in which case you won't see this isuse or you have this issue raised in your code, then I don't think having a list of exceptions will help the developer to fix it. If he thinks that it's a FP because we don't handle the framework it's using, then he can reach out to us and we will investigate the FP. If for example it's just plain HTML, having a list of exceptions for specific frameworks is just noise.


Be aware that the rule description has been synced in sonar-html  as part of a separate PR and it has been merged into master. See commit https://github.com/SonarSource/sonar-html/commit/a8ae2d33f11a2812f6f67867e5a3feb2798f2013

asya-vorobeva:
> Hi @asya-vorobeva thanks for the review, it's true that the rspec rule description is bad, but it was improved in this PR: [SonarSource/rspec#7179](https://github.com/SonarSource/rspec/pull/7179)
> 
> Though there is no exception listed in the rule description because I think that either you fall into one of the exception, in which case you won't see this isuse or you have this issue raised in your code, then I don't think having a list of exceptions will help the developer to fix it. If he thinks that it's a FP because we don't handle the framework it's using, then he can reach out to us and we will investigate the FP. If for example it's just plain HTML, having a list of exceptions for specific frameworks is just noise.
> 
> Be aware that the rule description has been synced in sonar-html as part of a separate PR and it has been merged into master. See commit [a8ae2d3](https://github.com/SonarSource/sonar-html/commit/a8ae2d33f11a2812f6f67867e5a3feb2798f2013)

If you check descriptions of other rules, you'll see that we usually provide exceptions in case they exist. I think that documentation about the rule should be full. It's not for users, it's for us developers (ordinary users even don't have an access to full RSpec descriptions).

francois-mora-sonarsource (commented):
The three issues look like cases where the false-positive remediation accepts too much weak evidence. Suggested simple scope-reduction criteria:

1. Razor tag-helper children in labels

   Only treat `<partial>` / `<vc:...>` as opaque generated label/control content when the analyzed file is Razor. Simple implementation: require `Helpers.isRazorFile(getHtmlSourceCode()) && Helpers.isRazorFragmentTagHelper(node)`.

2. `th:attr="text=..."` / `utext=...`

   Only count template-generated element text when the template engine attribute actually writes element body text. Simple implementation: accept direct `th:text`, `th:utext`, `v-text`, `v-html` with non-empty values; do not route `"text"` / `"utext"` through `Thymeleaf.hasNonEmptyThymeleafAttribute`, because `th:attr` only produces HTML attributes.

3. Empty Thymeleaf heading text

   Template text should count as heading content only if it would render non-empty text. Simple implementation: for `th:text` / `th:utext`, use `!Thymeleaf.isEmptyValue(attributeValue)` instead of `attributeValue != null && !attributeValue.isBlank()`; this rejects empty single-quoted literals, `""`, and whitespace-only quoted literals.

erwan-leforestier-sonarsource:
Addressed François's feedback and synced the rule description.

What changed:
- Gated Razor `<partial>` / `<vc:...>` label children so they are treated as opaque only in Razor files.
- Stopped treating `th:attr="text=..."` / `th:attr="utext=..."` as generated element body text, since `th:attr` only writes attributes.
- Reused `Thymeleaf.isEmptyValue(...)` for heading template text, so empty quoted literals like `''` and blank quoted literals no longer count as accessible heading content.
- Added regression coverage for those three cases.
- Regenerated `S6853.html` from the latest merged rspec `master` after SonarSource/rspec#7218, using the rule-api jar.

Validation:
- `mvn -pl sonar-html-plugin -Dtest=LabelHasAssociatedControlCheckTest,HeadingHasAccessibleContentCheckTest,AnchorsHaveContentCheckTest,ImgWithoutAltCheckTest,TableWithoutHeaderCheckTest test`
- `mvn -pl sonar-html-plugin test`
- `mvn -pl sonar-html-plugin -Dtest=LocalizedBundlesTest test`

francois-mora-sonarsource (approved):

PR #743, opened 2026-07-15
SONARHTML-409 Allow tabpanel with non-negative tabIndex

erwan-leforestier-sonarsource:
## Summary
This fixes S6845 false positives on HTML elements that use the `tabpanel` role with a non-negative `tabIndex`.
It also recognizes the same exception for static framework-bound role literals so the HTML rule aligns with the JS/TS variant.

## Changes
- Allow `tabpanel` as an exception for S6845 in the HTML check.
- Read static role bindings through `getPropertyValue("role")` so framework-bound `tabpanel` and interactive roles are handled consistently.
- Add regression coverage for compliant `tabpanel`/interactive role bindings and a noncompliant `tooltip` counterexample.

## Functional Validation
Artifact: [SONARHTML-409-fv.zip](https://github.com/user-attachments/files/30079671/SONARHTML-409-fv.zip)

    ./run.sh

Expected output:
```text
Analyzing "tabpanel-tabindex.html"...
Results:
    - Rule "S6845" -> L.1
    - Rule "S6845" -> L.2
    - Rule "S6845" -> L.3

****** Branch "fix/sonarhtml-409-allow-tabpanel-tabindex" ******
Analyzing "tabpanel-tabindex.html"...
Results:
    - Rule "S6845" -> L.3
```

### DIFF
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoNoninteractiveTabIndexCheck.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoNoninteractiveTabIndexCheck.java
@@ -57,13 +57,43 @@ public class NoNoninteractiveTabIndexCheck extends AbstractPageCheck {
    * @return true when the element declares an interactive role or the tabpanel role
    */
   private static boolean hasAllowedRole(TagNode node) {
-    var role = node.getPropertyValue("role");
+    var role = resolveStaticRole(node);
     if (role == null) {
       return false;
     }
 
-    var normalizedRole = role.replace("\"", "").replace("'", "").trim();
-    return "tabpanel".equalsIgnoreCase(normalizedRole)
-      || INTERACTIVE_ROLES.stream().anyMatch(normalizedRole::equalsIgnoreCase);
+    return "tabpanel".equalsIgnoreCase(role)
+      || INTERACTIVE_ROLES.stream().anyMatch(role::equalsIgnoreCase);
+  }
+
+  private static String resolveStaticRole(TagNode node) {
+    var roleAttribute = node.getProperty("role");
+    if (roleAttribute == null) {
+      return null;
+    }
+
+    var roleValue = roleAttribute.getValue();
+    if (roleAttribute.getName().equalsIgnoreCase("role")) {
+      var normalizedRole = roleValue.trim();
+      return normalizedRole.isEmpty() ? null : normalizedRole;
+    }
+
+    // Bound role values are only trusted when they are explicit string literals.
+    if (!isStaticStringLiteral(roleValue)) {
+      return null;
+    }
+
+    var normalizedRole = roleValue.substring(1, roleValue.length() - 1).trim();
+    return normalizedRole.isEmpty() ? null : normalizedRole;
+  }
+
+  private static boolean isStaticStringLiteral(String value) {
+    if (value == null || value.length() < 2) {
+      return false;
+    }
+
+    char first = value.charAt(0);
+    char last = value.charAt(value.length() - 1);
+    return (first == '\'' && last == '\'') || (first == '"' && last == '"');
   }
 }
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoNoninteractiveTabIndexCheck.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoNoninteractiveTabIndexCheck.java
@@ -88,7 +88,7 @@ public class NoNoninteractiveTabIndexCheck extends AbstractPageCheck {
   }
 
   private static boolean isStaticStringLiteral(String value) {
-    if (value == null || value.length() < 2) {
+    if (value.length() < 2) {
       return false;
     }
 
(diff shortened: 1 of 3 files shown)


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoNoninteractiveTabIndexCheck.java
lijun-chen-sonarsource: Perhaps out of the scope for this PR, but since in the accompanying rspec PR you've switch to using `tabindex` instead of `tabIndex` (which is indeed for common for HTML), perhaps also here the message could use `tabindex`.

Feel free to ignore.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoNoninteractiveTabIndexCheck.java
```diff
@@ -49,4 +49,21 @@ public void startElement(TagNode node) {
...
+   * @param node the element whose role declaration is inspected
+   * @return true when the element declares an interactive role or the tabpanel role
+   */
+  private static boolean hasAllowedRole(TagNode node) {
+    var role = node.getPropertyValue("role");
+    if (role == null) {
+      return false;
+    }
+
+    var normalizedRole = role.replace("\"", "").replace("'", "").trim();
```
lijun-chen-sonarsource: IIUC In trying to handle a FP for Vue.js, this introduces the false negative
```
<div :role="tabpanel" tabIndex="0" />
```
Here `tabpanel` is a variable name dynamically bound to `:role` .
Perhaps acceptable, but it's would be good to acknowledge it in the test.-ht

Also, since sonar-html supports all sorts of framework with dynamic binding, I guess this is a much larger concern and source of both FPs and FNs. I didn't find existing helper to deal with this case.
Most probably out of the scope for this PR, but perhaps worth creating a ticket for this.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoNoninteractiveTabIndexCheck.java
```diff
@@ -49,4 +49,21 @@ public void startElement(TagNode node) {
       // ignore
     }
   }
+
+  /**
+   * Returns whether the element declares a role that allows a non-negative tabindex.
+   *
+   * @param node the element whose role declaration is inspected
+   * @return true when the element declares an interactive role or the tabpanel role
```
lijun-chen-sonarsource, line 57: Strictly speaking, the WAI-ARIA tabs pattern says

> When the tabpanel does not contain any focusable elements or the first element with content is not focusable, the tabpanel should set tabindex="0" to include it in the tab sequence of the page.

So if the first element of the tabpanel is focusable, then we shouldn't allow tabindex here.
But probably it's not worth the effort.

Feel free to ignore this comment.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoNoninteractiveTabIndexCheck.java
```diff
@@ -49,4 +49,21 @@ public void startElement(TagNode node) {
...
+   */
+  private static boolean hasAllowedRole(TagNode node) {
```
lijun-chen-sonarsource, line 59: Do you think this `tabpanel` exception is general enough that we can put it in the shared `hasInteractiveRole` method? Or you think it is better to keep it local to this rule?

lijun-chen-sonarsource (changes requested):

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoNoninteractiveTabIndexCheck.java
erwan-leforestier-sonarsource: Not worth the effort. It would be nice to detect if the tabpanel contains focusable elements but when we have dynamic components, it's too difficult to evaluate this without making mistakes. Better to keep it simple and allow tabindex="0" for all elements that have role="tabpanel" set.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoNoninteractiveTabIndexCheck.java
erwan-leforestier-sonarsource: Tabpanel is not an interactive role so it cannot be added to the shared `hasInteractiveRole`. I prefer to keep it local to this rule.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoNoninteractiveTabIndexCheck.java
erwan-leforestier-sonarsource: Handled in 553c48a: bound role values are now trusted only when they are explicit string literals, so `:role="tabpanel"` is treated as dynamic and remains noncompliant. I also added that case to the test fixture.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoNoninteractiveTabIndexCheck.java
erwan-leforestier-sonarsource: You were right; the earlier reply on this thread was premature. This is now actually fixed in dfec0700: the rule message and test expectation both use `tabindex`.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoNoninteractiveTabIndexCheck.java
lijun-chen-sonarsource: Fair enough.

lijun-chen-sonarsource (approved):
LGTM!

PR #748, opened 2026-07-17
SONARHTML-412 Add regression test for S7930 Razor code-block preamble FP

erwan-leforestier-sonarsource:
## Summary

Adds a regression test for the `S7930` (NoDuplicateID) false positive described in SONARHTML-412: a Razor `@if / else if` chain preceded by a top-level `@{ ... }` code block (itself containing a C# `if`) reported duplicate ids on the mutually-exclusive branches.

The underlying bug was already fixed on `master` by the SONARHTML-411 rework (#744), which reworked `TemplateConditionalScopeTracker`. That PR did not, however, include a test for this specific shape, so the fix was unguarded and could silently regress. This change closes that gap.

## What this adds

- `conditionalBlocksRazorPreamble.cshtml` — the SONARHTML-412 repro: a top-level `@{ ... }` preamble followed by an `@if / else if / else if` chain reusing `id="home_vhead"`, plus a genuine duplicate `id="footer"` outside any conditional.
- `razorConditionalBlocksWithCodeBlockPreamble` — asserts only the real outside duplicate is flagged; the three branch ids are not.

## Validation

- Passes on current `master`.
- Verified as a real guard: swapping in the pre-fix `TemplateConditionalScopeTracker` makes the test fail (the branch false positives reappear).

No production code changes — test-only.

### DIFF
--- a/sonar-html-plugin/src/test/java/org/sonar/plugins/html/checks/coding/NoDuplicateIDCheckTest.java
+++ b/sonar-html-plugin/src/test/java/org/sonar/plugins/html/checks/coding/NoDuplicateIDCheckTest.java
@@ -241,6 +241,19 @@ class NoDuplicateIDCheckTest {
   }
 
   @Test
+  void razorConditionalBlocksWithCodeBlockPreamble() {
+    HtmlSourceCode sourceCode = TestHelper.scan(
+        new File("src/test/resources/checks/NoDuplicateIDCheck/conditionalBlocksRazorPreamble.cshtml"),
+        new NoDuplicateIDCheck());
+
+    // A top-level @{ ... } code block containing a C# if must not consume the conditional depth of the
+    // following @if/else if chain; ids across those branches stay mutually exclusive
+    checkMessagesVerifier.verify(sourceCode.getIssues())
+        .next().atLine(27).withMessage("Duplicate id \"footer\" found. First occurrence was on line 26.")
+        .noMore();
+  }
+
+  @Test
   void razorConditionalBlocksWithScriptTemplateLiteral() {
     HtmlSourceCode sourceCode = TestHelper.scan(
         new File("src/test/resources/checks/NoDuplicateIDCheck/conditionalBlocksScriptTemplateLiteral.cshtml"),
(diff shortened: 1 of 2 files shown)


sonar-html-plugin/src/test/java/org/sonar/plugins/html/checks/coding/NoDuplicateIDCheckTest.java
```diff
@@ -240,6 +240,19 @@ void razorConditionalBlocksWithElseIf() {
...
   }
 
+  @Test
+  void razorConditionalBlocksWithCodeBlockPreamble() {
+    HtmlSourceCode sourceCode = TestHelper.scan(
+        new File("src/test/resources/checks/NoDuplicateIDCheck/conditionalBlocksRazorPreamble.cshtml"),
+        new NoDuplicateIDCheck());
+
+    // A top-level @{ ... } code block containing a C# if must not consume the conditional depth of the
+    // following @if/else if chain; ids across those branches stay mutually exclusive
```
lijun-chen-sonarsource, line 250: > a C# if must not consume the conditional depth of the following @if/else if chain

I can't understand this comment. I guess someone with knowledge of the work done at SONARHTML-411 would understand this. A future reader would have to do a bit of work to connect the dots though.

Overall, I wonder whether we need to be so specific with implementation details in this comment, or whether something like "Ids across branches in @if/else if chain stay mutually exclusive, even if the chain is preceded by a code block preamble".
Or maybe we can make it more explicit that this is a regression test, for a fix done in SONARHTML-411.

Feel free to ignore this comment.

lijun-chen-sonarsource (approved):
LGTM!

I left a nitpick. But you can ignore it.

sonar-html-plugin/src/test/java/org/sonar/plugins/html/checks/coding/NoDuplicateIDCheckTest.java
erwan-leforestier-sonarsource: Applied your suggestion by simplifying the comment

PR #774, opened 2026-08-11
SONARHTML-438 Fix S6851 FP: false positives on unresolved alt bindings

victor-peixoto-sonarsource:
## What

`Web:S6851` (`ImgRedundantAltCheck`) flagged `img` `alt` attributes bound via
Angular (`[alt]`, `[attr.alt]`) or Vue (`:alt`, `v-bind:alt`) syntax whenever
the *binding expression* happened to contain "image", "photo", or "picture" —
even though the actual rendered text is unknown at analysis time.

```html
<img [src]="data.imageSrc" [alt]="data.imageAlt">   <!-- was FP -->
<img :src="imageSrc" :alt="IMAGE_ALT">              <!-- was FP -->
```

## Why

Customer feedback surfaced 7 grouped reports (11 occurrences across 5
projects) of this exact pattern, and the local ruling corpora contain ~42
similar dynamic `alt` binding lines across 37 files.

`TagNode.getProperty("alt")` already resolves all the Angular/Vue binding
forms to find the attribute, but the check was calling `getPropertyValue`
and scanning whatever came back — including raw JS/TS expressions — as if it
were literal alt text.

## Fix

- `ImgRedundantAltCheck` now looks at the original attribute name returned
by `getProperty("alt")`:
  - Plain `alt="..."` — unchanged behavior.
  - A binding form (`[alt]`, `[attr.alt]`, `attr.alt`, `:alt`, `v-bind:alt`,
`:[alt]`) — only inspected when the value is a statically resolvable JS
string literal (`[alt]="'image of a sunrise'"`); any other expression is
treated as unresolved and skipped.
- Added `AccessibilityUtils.unwrapStaticStringLiteral`, lifting the
string-literal-unwrap logic that already existed privately in
`ControlGroup` (used for `input[type]` bindings) so it's shared rather than
duplicated a third time. (`ControlGroup` itself is left untouched — no
behavior change there, out of scope for this fix.)

## Testing

- New fixture `ImgRedundantAltCheck/bound-alt.html` covers every supported
Angular/Vue binding form with both a noncompliant string literal and an
unresolved expression, plus the existing `static/aria-hidden` cases.
- New `boundAlt()` test in `ImgRedundantAltCheckTest`.
- Existing `test()` (plain `alt=` cases) is unchanged and still passes.
- Full `sonar-html-plugin` test module passes.

### DIFF
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/AccessibilityUtils.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/AccessibilityUtils.java
@@ -17,6 +17,7 @@
 package org.sonar.plugins.html.api.accessibility;
 
 import java.util.Set;
+import javax.annotation.CheckForNull;
 import org.sonar.plugins.html.api.Thymeleaf;
 import org.sonar.plugins.html.node.TagNode;
 
@@ -66,6 +67,26 @@ public class AccessibilityUtils {
 
     var ariaDisabledAttr = element.getAttribute("aria-disabled");
     return "true".equalsIgnoreCase(ariaDisabledAttr);
+  }
+
+  /**
+   * Unwraps a JavaScript string literal used as a framework binding value
+   * (Angular {@code [x]="'foo'"}, Vue {@code :x="'foo'"}), returning the inner
+   * text. Returns {@code null} when {@code value} is not a single- or
+   * double-quoted string literal — i.e. it's a dynamic expression that cannot
+   * be statically resolved.
+   */
+  @CheckForNull
+  public static String unwrapStaticStringLiteral(@CheckForNull String value) {
+    if (value == null || value.length() < 2) {
+      return null;
+    }
+    char first = value.charAt(0);
+    char last = value.charAt(value.length() - 1);
+    if ((first == '\'' && last == '\'') || (first == '"' && last == '"')) {
+      return value.substring(1, value.length() - 1);
+    }
+    return null;
   }
 
   public static boolean isFocusableElement(TagNode element) {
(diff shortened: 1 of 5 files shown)


sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/AccessibilityUtils.java
```diff
@@ -68,6 +70,61 @@ public static boolean isDisabledElement(TagNode element) {
...
+        // means the expression is more than this one literal.
+        return index == expression.length() - 1 ? expression.substring(1, index) : null;
+      } else {
+        index++;
+      }
+    }
+
+    // The opening quote is never closed.
+    return null;
+  }
```
erwan-leforestier-sonarsource, line 116: [OPT] The `getProperty` + unwrap approach is good, but `unwrapStaticStringLiteral` is now the 4th copy of the "strip quotes off a bound value" logic. `ControlGroup` and `NoNoninteractiveTabIndexCheck` both have their own `isStaticStringLiteral` that only checks the first and last char, so they misread things like `[role]="'a' + x + 'b'"` as a literal, a bug your new version actually fixes. Could you migrate those two onto `unwrapStaticStringLiteral` and drop their private copies? Same idea for the binding-form check: `"alt".equalsIgnoreCase(name)` is the same `isBindingForm` helper already duplicated in `InputWithoutLabelCheck` and `LabelHasAssociatedControlCheck`, so it would be nice to share one.

erwan-leforestier-sonarsource (approved):

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/AccessibilityUtils.java
victor-peixoto-sonarsource: Yes, thanks for raising that. I had this ticket for ControlGroup already. https://sonarsource.atlassian.net/browse/SONARHTML-444 
I will add NoNoninteractiveTabIndexCheck to it too. I will do it in a separate PR after merging this one.

PR #776, opened 2026-08-12
SONARHTML-436 Accept div groups in description lists

erwan-leforestier-sonarsource:
## Summary
S1093 now recognizes valid description-list groups whose `dt` elements are directly inside a `div` child of `dl`, avoiding false positives on valid markup.

## Changes
- Accept a `div` direct child of `dl` as an allowed parent for `dt`
- Keep nested `div` structures reported
- Cover single and multiple term/definition groups in the rule fixtures

## Functional Validation
Artifact: [SONARHTML-436-fv.zip](https://github.com/user-attachments/files/30976710/SONARHTML-436-fv.zip)

    ./run.sh

Expected output:
```text
Analyzing "sample.html"...
Results:
    - Rule "Web:S1093" -> L.3
    - Rule "Web:S1093" -> L.8

****** Branch "fix/sonarhtml-436-accept-dl-div-groups" ******
Analyzing "sample.html"...
Results:
    - Rule "Web:S1093" -> L.8
```

### DIFF
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/ItemTagNotWithinContainerTagCheck.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/ItemTagNotWithinContainerTagCheck.java
@@ -46,7 +46,7 @@ public class ItemTagNotWithinContainerTagCheck extends AbstractPageCheck {
     TagNode parent = effectiveParent(node);
     if (isLi(node) && shouldReport(parent, ItemTagNotWithinContainerTagCheck::isLiAllowedParent)) {
       createViolation(node, "Surround this <" + node.getNodeName() + "> item tag by a <ul>, <ol> or <menu> container one.");
-    } else if (isDt(node) && shouldReport(parent, ItemTagNotWithinContainerTagCheck::isDl)) {
+    } else if (isDt(node) && shouldReport(parent, ItemTagNotWithinContainerTagCheck::isDtAllowedParent)) {
       createViolation(node, "Surround this <" + node.getNodeName() + "> item tag by a <dl> container one.");
     }
   }
@@ -120,6 +120,26 @@ public class ItemTagNotWithinContainerTagCheck extends AbstractPageCheck {
 
   private static boolean isDl(TagNode node) {
     return "DL".equalsIgnoreCase(node.getNodeName());
+  }
+
+  /**
+   * Checks whether a node can directly contain a {@code dt} element.
+   *
+   * @param node the direct parent of the {@code dt} element
+   * @return true when the parent is a {@code dl} or a {@code div} directly inside a {@code dl}
+   */
+  private static boolean isDtAllowedParent(TagNode node) {
+    return isDl(node) || (isDiv(node) && node.getParent() != null && isDl(node.getParent()));
+  }
+
+  /**
+   * Checks whether a node is a {@code div} element.
+   *
+   * @param node the node to inspect
+   * @return true when the node is a {@code div} element
+   */
+  private static boolean isDiv(TagNode node) {
+    return "DIV".equalsIgnoreCase(node.getNodeName());
   }
 
   private static boolean isP(TagNode node) {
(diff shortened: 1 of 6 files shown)


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/ItemTagNotWithinContainerTagCheck.java
```diff
@@ -122,6 +122,26 @@ private static boolean isDl(TagNode node) {
...
   }
 
+  /**
+   * Checks whether a node can directly contain a {@code dt} element.
+   *
+   * @param node the direct parent of the {@code dt} element
+   * @return true when the parent is a {@code dl} or a {@code div} directly inside a {@code dl}
+   */
+  private static boolean isDtAllowedParent(TagNode node) {
+    return isDl(node) || (isDiv(node) && node.getParent() != null && isDl(node.getParent()));
```
victor-peixoto-sonarsource, line 132: Suggestion to be followed up either here or in a separate PR.

The ticket expected behavior implies we cover cases similar to [these](https://github.com/SonarSource/sonar-html/blob/0bdeea37553f577ab9afdd35df438a15b685b2ba/sonar-html-plugin/src/test/resources/checks/ItemTagNotWithinContainerTagCheck/valid-cases.html#L104-L114). But this snippet is still an FP:

```html
<dl>
  <c:forEach items="${items}" var="i">
    <div>
      <dt>${i.term}</dt>
      <dd>${i.def}</dd>
    </div>
  </c:forEach>
</dl>
```

A possible fix is to change this method to be:

```java
private static boolean isDtAllowedParent(TagNode node) {
  if (isDl(node)) {
    return true;
  }
  if (!isDiv(node)) {
    return false;
  }
  // Skip framework wrappers (c:forEach, th:block, …): they hide the real structure
  TagNode ancestor = node.getParent();
  while (ancestor != null && !hasKnownHTMLTag(ancestor)) {
    ancestor = ancestor.getParent();
  }
  return ancestor != null && isDl(ancestor);
}
```

Note: current `hasKnownHTMLTag` lookup is O(n). The fix would be to change the known tags list to a set https://github.com/SonarSource/sonar-html/blob/ee59a7a8a3b15c022e27ef4b01687d8c5fe70a4f/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/HtmlConstants.java#L50

sonar-html-plugin/src/test/resources/checks/ItemTagNotWithinContainerTagCheck/invalid-cases.html
victor-peixoto-sonarsource, line 35: Worth adding this

```suggestion
</div>

<div>
  <dt>term</dt><!-- Non-Compliant - a div outside any dl is not a valid parent -->
</div>
```

victor-peixoto-sonarsource: Maybe this too 
```suggestion
</div>

<dl>
  <div>
    <div>
      <dt>z</dt><!-- Non-Compliant - the inner div is not a direct child of the dl -->
    </div>
  </div>
</dl>
```

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/ItemTagNotWithinContainerTagCheck.java
victor-peixoto-sonarsource, line 132: Maybe for a different ticket, but `<dd>` should have the same behavior as per https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dd#:~:text=A%20%3Cdl%3E%20or%20a%20%3Cdiv%3E%20that%20is%20a%20child%20of%20a%20%3Cdl%3E.%0AThis%20element%20can%20be%20used%20after%20a%20%3Cdt%3E%20or%20another%20%3Cdd%3E%20element.
This method can change to be `isDefinitionItemAllowedParent`.

victor-peixoto-sonarsource (approved):
LGTM. Added some comments for possible follow-ups.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/ItemTagNotWithinContainerTagCheck.java
erwan-leforestier-sonarsource: Using `!hasKnownHTMLTag` would introduce many FNs for custom components. It's tradeoff, here I prefer to introduce FPs instead of FNs since I think we will have less FPs than FNs. The other solution is to detect framework wrappers that are not rendering intermediate elements in the DOM but it's kind of  a nightmare to maintain imo. I prefer to keep it as is.

erwan-leforestier-sonarsource: Added support for `dd` in commit 9111ff0c19e3b8ac33d50dc34fc1a42cc94ec1e3

PR #777, opened 2026-08-12
SONARHTML-441 Restrict S1082 to native HTML elements

erwan-leforestier-sonarsource:
## Summary
S1082 now evaluates mouse-event keyboard equivalence only for known native HTML elements, avoiding false positives on framework and design-system components. A native-element allowlist remains available for explicit opt-outs, and Vue mouse-event shorthands are recognized.
RSPEC PR: https://github.com/SonarSource/rspec/pull/7862

## Changes
- Gate S1082 before role and mouse-event evaluation using the canonical native tag classifier.
- Retain `whitelistedElements` for native-element opt-outs and document its full scope.
- Cover Angular, Vue, and native HTML mouse-handler cases.

## Functional Validation
Artifact:  [SONARHTML-441-fv.zip](https://github.com/user-attachments/files/30980438/SONARHTML-441-fv.zip)

    ./run.sh

Expected output:
```text
Analyzing "sample.html"...
Results:
    - Rule "Web:MouseEventWithoutKeyboardEquivalentCheck" -> L.1
    - Rule "Web:MouseEventWithoutKeyboardEquivalentCheck" -> L.2
    - Rule "Web:MouseEventWithoutKeyboardEquivalentCheck" -> L.3
    - Rule "Web:MouseEventWithoutKeyboardEquivalentCheck" -> L.4
    - Rule "Web:MouseEventWithoutKeyboardEquivalentCheck" -> L.5
    - Rule "Web:MouseEventWithoutKeyboardEquivalentCheck" -> L.6
    - Rule "Web:MouseEventWithoutKeyboardEquivalentCheck" -> L.7
    - Rule "Web:MouseEventWithoutKeyboardEquivalentCheck" -> L.8

****** Branch "fix/sonarhtml-441-native-html-s1082" ******
Analyzing "sample.html"...
Results:
    - Rule "Web:MouseEventWithoutKeyboardEquivalentCheck" -> L.1
    - Rule "Web:MouseEventWithoutKeyboardEquivalentCheck" -> L.2
    - Rule "Web:MouseEventWithoutKeyboardEquivalentCheck" -> L.3
    - Rule "Web:MouseEventWithoutKeyboardEquivalentCheck" -> L.11
    - Rule "Web:MouseEventWithoutKeyboardEquivalentCheck" -> L.12
    - Rule "Web:MouseEventWithoutKeyboardEquivalentCheck" -> L.13
```

### DIFF
--- a/its/ruling/src/test/resources/expected/Web-MouseEventWithoutKeyboardEquivalentCheck.json
+++ b/its/ruling/src/test/resources/expected/Web-MouseEventWithoutKeyboardEquivalentCheck.json
@@ -674,9 +674,6 @@
 108,
 130
 ],
-"project:voten/resources/assets/js/components/passport/PersonalAccessTokens.vue": [
-101
-],
 "project:voten/resources/assets/js/components/submission/GifSubmission.vue": [
 19
 ],
--- a/its/ruling/src/test/resources/expected/Web-MouseEventWithoutKeyboardEquivalentCheck.json
+++ b/its/ruling/src/test/resources/expected/Web-MouseEventWithoutKeyboardEquivalentCheck.json
@@ -564,11 +564,126 @@
 "project:sonar-master/sonar-server/src/main/webapp/WEB-INF/app/views/resource/_options.html.erb": [
 26
 ],
+"project:voten/resources/assets/js/components/AdminPanelSuggestedChannels.vue": [
+85
+],
+"project:voten/resources/assets/js/components/Announcement.vue": [
+11
+],
+"project:voten/resources/assets/js/components/BannedUser.vue": [
+13,
+18
+],
+"project:voten/resources/assets/js/components/BlockedDomain.vue": [
+12,
+18
+],
+"project:voten/resources/assets/js/components/BookmarkedChannel.vue": [
+25,
+33
+],
+"project:voten/resources/assets/js/components/BookmarkedUser.vue": [
+21,
+27
+],
+"project:voten/resources/assets/js/components/ChannelHeader.vue": [
+70,
+104
+],
+"project:voten/resources/assets/js/components/Comment.vue": [
+74,
+110,
+119
+],
+"project:voten/resources/assets/js/components/CommentForm.vue": [
+15,
+85
+],
+"project:voten/resources/assets/js/components/EmojiPicker.vue": [
+18,
+25,
+40
+],
 "project:voten/resources/assets/js/components/GifPlayer.vue": [
 22
 ],
+"project:voten/resources/assets/js/components/Home.vue": [
+24
+],
+"project:voten/resources/assets/js/components/Message.vue": [
+15
+],
+"project:voten/resources/assets/js/components/Messages.vue": [
+54,
+87,
+99,
+128,
+148,
+260
+],
+"project:voten/resources/assets/js/components/Moderator.vue": [
+17
+],
+"project:voten/resources/assets/js/components/NewSubmission.vue": [
+193,
+202,
+211
+],
+"project:voten/resources/assets/js/components/Notification.vue": [
+2
+],
+"project:voten/resources/assets/js/components/PhotoViewer.vue": [
+37
+],
+"project:voten/resources/assets/js/components/QuickChannelPicker.vue": [
+20
+],
+"project:voten/resources/assets/js/components/QuickEmojiPicker.vue": [
+18
+],
+"project:voten/resources/assets/js/components/QuickMentioner.vue": [
+20
+],
+"project:voten/resources/assets/js/components/ReportedComment.vue": [
+46,
+55,
+64
+],
+"project:voten/resources/assets/js/components/ReportedSubmission.vue": [
+43,
+49,
+55
+],
+"project:voten/resources/assets/js/components/Rule.vue": [
+10,
+14
+],
+"project:voten/resources/assets/js/components/Submission.vue": [
+161,
+173
+],
+"project:voten/resources/assets/js/components/UserHeader.vue": [
+127
+],
+"project:voten/resources/assets/js/components/auth/RightSidebar.vue": [
+9,
+41,
+50,
+52,
+106,
+108,
+130
+],
+"project:voten/resources/assets/js/components/passport/PersonalAccessTokens.vue": [
+101
+],
 "project:voten/resources/assets/js/components/submission/GifSubmission.vue": [
 19
+],
+"project:voten/resources/assets/js/components/submission/ImgSubmission.vue": [
+13,
+25,
+37
 ],
 "project:voten/resources/views/passport/index.blade.php": [
 26,
--- a/its/ruling/src/test/resources/expected/Web-MouseEventWithoutKeyboardEquivalentCheck.json
+++ b/its/ruling/src/test/resources/expected/Web-MouseEventWithoutKeyboardEquivalentCheck.json
@@ -3,41 +3,12 @@
 259,
 262
 ],
-"project:Silverpeas-Core-master/war-core/src/main/webapp/form/jsp/editor.jsp": [
-503,
-506
-],
-"project:Silverpeas-Core-master/war-core/src/main/webapp/jobDomainPeas/jsp/userCreate.jsp": [
-238
-],
 "project:Silverpeas-Core-master/war-core/src/main/webapp/jobDomainPeas/jsp/userImport.jsp": [
 201,
 209
 ],
-"project:Silverpeas-Core-master/war-core/src/main/webapp/pdcPeas/jsp/axisManager.jsp": [
-283,
-284
-],
-"project:Silverpeas-Core-master/war-core/src/main/webapp/pdcPeas/jsp/globalResult.jsp": [
-593
-],
-"project:Silverpeas-Core-master/war-core/src/main/webapp/pdcPeas/jsp/searchContext.jsp": [
-411,
-557
-],
-"project:Silverpeas-Core-master/war-core/src/main/webapp/pdcPeas/jsp/searchContextInComponent.jsp": [
-550,
-710
-],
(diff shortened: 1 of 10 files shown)


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/MouseEventWithoutKeyboardEquivalentCheck.java
```diff
@@ -165,8 +161,9 @@ private static boolean hasEventHandlerAttribute(TagNode node, String eventName)
       || hasAttribute(node, "(" + eventName + ")")
       || hasAttribute(node, "ON-" + eventName)
       || hasAttribute(node, "NG-" + eventName)
-      // Vue long form: v-on:eventname (shorthand @eventname has @ stripped by the parser)
-      || hasAttribute(node, "V-ON:" + eventName);
+      // Vue long form v-on:eventname, plus the @eventname shorthand the parser strips to a bare name
+      || hasAttribute(node, "V-ON:" + eventName)
+      || hasAttribute(node, eventName);
```
victor-peixoto-sonarsource: This line makes bare event names count as handlers for every event, but modifiers are still only
understood by the two `KEY_*_WITH_KEY_NAME` patterns. That asymmetry cuts both ways — a modifier on
the *satisfying* side hides a real handler, and one on the *triggering* side hides the trigger:

| Case | Current behaviour |
| --- | --- |
| `<div @mouseover="a" @focus.capture="b">` | raises `Add a 'onFocus'…` — `@focus.capture` not recognised |
| `<div @mouseout="a" @blur.once="b">` | raises `Add a 'onBlur'…` — `@blur.once` not recognised |
| `<div @click="a" @keydown.arrow-up="b">` | raises — `\w{1,10}` rejects Vue's kebab-case key modifiers |
| `<div @click.prevent="a">` | raises nothing — `@click.prevent` not recognised as a click handler |
| `<div v-on:click.prevent="a">` | raises nothing — same |

Generalising the modifier grammar across all events fixes all five. The validation in the existing
patterns has to be preserved though — key names capped at 10 characters and 5 combined modifiers is
what keeps `(keydown.enter,shift)` and `(keydown.undefinedkeyname)` correctly reported (fixture lines
68, 70, 71). A naive "strip everything after the first dot" accepts those and breaks `detected()`.

```java
// Event modifiers / key names: Angular's (keydown.enter) pseudo-events and Vue's @click.prevent,
// @keydown.arrow-up. Key names are capped at 10 characters and 5 combined modifiers so that malformed
// bindings — (keydown.enter,shift), (keydown.undefinedkeyname) — are not mistaken for real handlers.
private static final String MODIFIERS = "(?:\\.[\\w-]{1,10}){0,5}";
private static final Map<String, Pattern> EVENT_PATTERNS = new ConcurrentHashMap<>();

private static boolean hasEventHandlerAttribute(TagNode node, String eventName) {
  return hasAttribute(node, "ON" + eventName)
    || hasAttribute(node, "ON-" + eventName)
    || hasAttribute(node, "NG-" + eventName)
    // Angular (eventname), Vue long form v-on:eventname, and the @eventname shorthand the parser
    // strips to a bare name — each optionally carrying modifiers (.prevent, .enter, .arrow-up)
    || hasEventBinding(node, eventName);
}

private static boolean hasEventBinding(TagNode node, String eventName) {
  Pattern pattern = EVENT_PATTERNS.computeIfAbsent(eventName, name -> Pattern.compile(
    "\\(" + name + MODIFIERS + "\\)"
      + "|(?:v-on:)?" + name + MODIFIERS,
    Pattern.CASE_INSENSITIVE));
  return node.getAttributes().stream().anyMatch(a -> pattern.matcher(a.getName()).matches());
}
```

This subsumes `KEY_DOWN_WITH_KEY_NAME`, `KEY_UP_WITH_KEY_NAME`, `hasKeyDownWithKeyName`,
`hasKeyUpWithKeyName` and the `(eventName)` branch, so all five can go — net −13 lines.

Applied locally at this branch's head: all five rows above behave correctly, the malformed key-name
cases stay reported, `<a href="#" @click.prevent="x">` stays exempt, and the full module passes
(**657 tests, 0 failures**, this check's four tests unchanged).

It also looks ruling-neutral: of the 52 `@click.<modifier>` sites in `its/sources`, 40 are on
`<el-dropdown-item>` (custom element, excluded by the new gate) and 12 on `<a @click.prevent>`
(exempt via `isHyperlink` — and that exemption only engages properly once the modifier form is
recognised). Worth a ruling run to confirm, but no churn expected on top of the +59.

Suggest adding those five rows as fixture cases either way; none are covered today.

its/ruling/src/test/resources/expected/Web-MouseEventWithoutKeyboardEquivalentCheck.json
```diff
@@ -564,12 +564,127 @@
...
+9,
+41,
+50,
+52,
+106,
+108,
+130
+],
+"project:voten/resources/assets/js/components/passport/PersonalAccessTokens.vue": [
+101
```
victor-peixoto-sonarsource: This entry is a false positive. `PersonalAccessTokens.vue:101` is:

```html
<input type="checkbox" @click="toggleScope(scope.id)" :checked="scopeIsAssigned(scope.id)">
```

A native checkbox is keyboard-operable by definition — Space activates it and the HTML activation
behaviour fires a synthetic `click`, so the `@click` handler *is* the keyboard path. Asking for
`onKeyDown`/`onKeyUp` here is wrong.

The cause is `isAnInteractiveRole`, which contains only `"textbox"`, so `<textarea>` and text inputs
are exempted via their implicit role while `checkbox`, `radio`, `combobox` and `listbox` fall through:

| Input | Result |
| --- | --- |
| `<input type="checkbox" @click="x">` | flagged (FP) |
| `<input type="checkbox" onclick="x">` | flagged — pre-existing on master |
| `<input type="radio" @click="x">` | flagged (FP) |
| `<select @click="x">` | flagged (debatable) |
| `<textarea @click="x">` | clean (textbox role) |

Since `onclick=` already behaves this way the gap predates this PR — the Vue support just gives it a
first corpus hit. Either widen `isAnInteractiveRole` to the natively-activatable roles, or split it
into its own ticket; what I'd avoid is accepting this line into the expectation file as though it
were correct.

For the record, the other 58 additions all check out — `<i>`, `<div>`, `<li>`, `<span>`, `<h1>`,
`` with `@click` and no keyboard equivalent, no `role`, no `tabindex`. I reconstructed each
flagged tag from the pinned corpus commit `ef27916` to confirm.

victor-peixoto-sonarsource (changes requested):
Good job. I left some comments

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/MouseEventWithoutKeyboardEquivalentCheck.java
erwan-leforestier-sonarsource: Handled in 33ee23f0. Event bindings now accept valid Angular/Vue modifiers consistently for every event, while retaining the existing modifier count and key-name length limits. The fixture covers all five cases from the review.

its/ruling/src/test/resources/expected/Web-MouseEventWithoutKeyboardEquivalentCheck.json
erwan-leforestier-sonarsource: Handled in 33ee23f0. Native implicit checkbox, radio, and listbox controls are now exempted; explicit ARIA roles still require keyboard handlers. Added native-control regression cases and removed the PersonalAccessTokens false positive from the ruling expectation.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/MouseEventWithoutKeyboardEquivalentCheck.java
```diff
@@ -33,31 +36,17 @@
...
-    "|keyup(\\.\\w{1,10}){1,5}" +
-    "|v-on:keyup(\\.\\w{1,10}){1,5}",
-    Pattern.CASE_INSENSITIVE);
+  private static final String DEFAULT_WHITELISTED_ELEMENTS = "";
+
+  // Angular pseudo-events and Vue modifiers. Key names are limited to 10 characters and realistic
+  // combinations to five modifiers so malformed bindings are not mistaken for event handlers.
+  private static final String EVENT_MODIFIERS = "(?:\\.[\\w-]{1,10}){0,5}";
+  private static final Map<String, Pattern> EVENT_PATTERNS = new ConcurrentHashMap<>();
+  private static final Set<String> NATIVELY_ACTIVATABLE_ROLES = Set.of("textbox", "checkbox", "radio", "listbox");
```
victor-peixoto-sonarsource: Do we also need combobox?

victor-peixoto-sonarsource (approved):
LGTM. I left one question.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/MouseEventWithoutKeyboardEquivalentCheck.java
erwan-leforestier-sonarsource: Good catch. A native single-select has an implicit combobox role, while multiple or size > 1 selects have listbox. I corrected the implicit-role mapping, included combobox among the natively keyboard-activatable roles, and added coverage for both forms. I also updated the rule description in RSPEC PR #7862 to cover the implicit combobox role and its listbox variants.

erwan-leforestier-sonarsource: Follow-up: I also updated the rule description in RSPEC PR #7862 to cover the implicit combobox role and its listbox variants.

PR #778, opened 2026-08-12
SONARHTML-442 Recognize labeled Angular Material fields

erwan-leforestier-sonarsource:
## Summary
Avoids S1097 false positives for Angular Material controls whose enclosing `mat-form-field` supplies a `mat-label`, including labels in structural descendants such as `ng-container`. Controls without that exact structural label relationship remain reportable.

## Changes
- Recognize `matInput` controls inside labeled Angular Material form fields
- Support direct, property-bound, and descendant `mat-label` forms
- Preserve findings for unlabeled fields, native controls, and unrelated components

## Functional Validation
Artifact: [SONARHTML-442-fv.zip](https://github.com/user-attachments/files/30979817/SONARHTML-442-fv.zip)

    ./run.sh

Expected output:
```text
Analyzing "sample.html"...
Results:
    - Rule "Web:InputWithoutLabelCheck" -> L.3
    - Rule "Web:InputWithoutLabelCheck" -> L.7
    - Rule "Web:InputWithoutLabelCheck" -> L.15

****** Branch "fix/sonarhtml-442-recognize-labeled-material-fields" ******
Analyzing "sample.html"...
Results:
    (none)
```

### DIFF
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/InputWithoutLabelCheck.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/InputWithoutLabelCheck.java
@@ -151,11 +151,21 @@ public class InputWithoutLabelCheck extends AbstractPageCheck {
     TagNode parent = node.getParent();
     while (parent != null) {
       if (parent.equalsElementName("mat-form-field")) {
-        return parent.getChildren().stream().anyMatch(child -> child.equalsElementName("mat-label"));
+        return containsMatLabel(parent);
       }
       parent = parent.getParent();
     }
     return false;
+  }
+
+  /**
+   * Determines whether an element contains an Angular Material label among its descendants.
+   * @param node the element whose descendants are inspected
+   * @return true when a mat-label is found
+   */
+  private static boolean containsMatLabel(TagNode node) {
+    return node.getChildren().stream()
+      .anyMatch(child -> child.equalsElementName("mat-label") || containsMatLabel(child));
   }
 
   /**
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/InputWithoutLabelCheck.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/InputWithoutLabelCheck.java
@@ -43,6 +43,8 @@ public class InputWithoutLabelCheck extends AbstractPageCheck {
   private static final String ASSOCIATE_LABEL_MESSAGE = "Associate a valid label to this input field.";
   private static final String ID = "id";
   private static final String ASP_FOR = "asp-for";
+  private static final Set<String> MAT_INPUT_ATTRIBUTE_NAMES = Set.of("matInput", "[matInput]");
+  private static final Set<String> MAT_NATIVE_CONTROL_ATTRIBUTE_NAMES = Set.of("matNativeControl", "[matNativeControl]");
 
   // LinkedHashMap/LinkedHashSet so endDocument emits issues in document order — matches tests and ruling reports.
   private final Set<String> labelTargets = new LinkedHashSet<>();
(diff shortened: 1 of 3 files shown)


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/InputWithoutLabelCheck.java
```diff
@@ -138,6 +138,47 @@ private static boolean isInputRequiredLabel(TagNode node) {
...
+  /**
+   * Determines whether a control declares Angular Material's matInput directive.
+   * @param node the control to inspect
+   * @return true when the directive is declared directly or through Angular property binding
+   */
+  private static boolean hasMatInputDirective(TagNode node) {
+    return node.getAttributes().stream()
+      .map(Attribute::getName)
+      .anyMatch(name -> "matInput".equalsIgnoreCase(name) || "[matInput]".equalsIgnoreCase(name));
+  }
```
victor-peixoto-sonarsource, line 198: Suggestion:

```java
private static final Set<String> MAT_CONTROL_DIRECTIVES = Set.of("matInput", "matNativeControl");

private static boolean hasMatInputDirective(TagNode node) {
  return MAT_CONTROL_DIRECTIVES.stream()
    .flatMap(directive -> Stream.concat(Stream.of(directive), TagNode.domPropertyBindingNames(directive).stream()))
    .anyMatch(spelling -> node.getAttributes().stream().anyMatch(a -> spelling.equalsIgnoreCase(a.getName())));
}

// stop at a nested field: its mat-label labels its own control, not this one
private static boolean containsMatLabel(TagNode node) {
  return node.getChildren().stream()
    .anyMatch(child -> child.equalsElementName("mat-label")
      || (!child.equalsElementName("mat-form-field") && containsMatLabel(child)));
}
```

**`MAT_CONTROL_DIRECTIVES` (adding `matNativeControl`)** — Angular Material's `MatInput` selector is
`input[matInput], textarea[matInput], select[matNativeControl], input[matNativeControl], textarea[matNativeControl]`.
Because `select[matInput]` is not part of it, `matNativeControl` is the *only* valid spelling for a
native `<select>` in a form field. Running the check at this branch's head, `<select matNativeControl>`
and `<input matNativeControl>` inside a labeled `mat-form-field` still raise — so that population keeps
the exact false positive this ticket fixes. Sourcing the spellings from `TagNode.domPropertyBindingNames`
(#779) also replaces the hand-rolled `[matInput]` string match with the shared definition that
SONARHTML-444 introduced for this purpose; the branch predates #779 and can pick it up on rebase.

**Boundary in `containsMatLabel`** — the recursion currently accepts any `mat-label` anywhere below the
field, so a label belonging to a *nested* `mat-form-field` satisfies the outer field's unlabeled control
(verified: an outer `<input matInput>` with no label of its own is silently accepted). Skipping nested
`mat-form-field` subtrees keeps the descendant support that was asked for while matching the PR
description's "that exact structural label relationship" — a nested field's `mat-label` labels its own
control, not this one.

Both are false-negative-direction issues on structurally invalid markup, so neither is urgent; the
`matNativeControl` gap is the one with real user impact.

**One trade-off to decide:** `domPropertyBindingNames` also returns the Vue spellings (`:matInput`,
`v-bind:matInput`), which are inert for an Angular-only directive. Harmless, but if you would rather not
accept them, keep an explicit list and note why — right now the exclusion of Vue forms reads as
incidental rather than intentional.

victor-peixoto-sonarsource (approved):
LGTM. Left one suggestion.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/InputWithoutLabelCheck.java
erwan-leforestier-sonarsource: Thanks, addressed both points: added tag-aware matNativeControl support and scoped labels to their own mat-form-field, while keeping bindings Angular-only.

PR #807, opened 2026-08-28
SONARHTML-472 Handle plain C# conditionals in Razor code blocks

victor-peixoto-sonarsource:
Part of [SONARHTML-472](https://sonarsource.atlassian.net/browse/SONARHTML-472)

## Summary
- recognize plain C# if / else if / else and switch branches inside Razor code blocks
- preserve branch scope across nested braces, strings, line comments, block comments, and Razor comments
- ignore HTML-like tags in non-rendered Razor/C# content while keeping the behavior limited to Razor files
- add regression coverage for nested and chained conditionals, switch branches, comments, and genuine duplicates

## Testing
- mvn verify (689 tests)

[SONARHTML-472]: https://sonarsource.atlassian.net/browse/SONARHTML-472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/TemplateConditionalScopeTracker.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/TemplateConditionalScopeTracker.java
@@ -49,7 +49,6 @@ public final class TemplateConditionalScopeTracker {
   private static final Pattern CSHARP_BLOCK_START_PATTERN = Pattern.compile("(if|switch)\\s*\\(", Pattern.CASE_INSENSITIVE);
   private static final Pattern CSHARP_GENERIC_ARGUMENT_PATTERN = Pattern.compile("<\\s*[\\p{L}_][\\p{L}\\p{N}_.,:?\\[\\]\\s<>]*>");
   private static final int CSHARP_RAW_STRING_MIN_QUOTE_COUNT = 3;
-  private static final int CSHARP_EMPTY_RAW_STRING_QUOTE_COUNT = 6;
   // Angular block control flow branches: @case (value) { and @default { inside an @switch
   private static final Pattern ANGULAR_BRANCH_START_PATTERN = Pattern.compile("@(case|default)\\s*[({]", Pattern.CASE_INSENSITIVE);
   private static final Pattern PHP_DIRECTIVE_CONDITIONAL_START_PATTERN = Pattern.compile("(if|foreach|for)\\b", Pattern.CASE_INSENSITIVE);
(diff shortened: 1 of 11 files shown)


sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/TemplateConditionalScopeTracker.java
```diff
@@ -123,29 +269,294 @@ private boolean hasOpenConditionalScope() {
...
+    return verbatimCSharpString && current == csharpStringDelimiter
+      && state.index + 1 < text.length() && text.charAt(state.index + 1) == csharpStringDelimiter;
+  }
+
+  private void updateCSharpStringState(char current) {
+    if (escapedCSharpStringCharacter) {
+      escapedCSharpStringCharacter = false;
+    } else if (!verbatimCSharpString && current == '\\') {
+      escapedCSharpStringCharacter = true;
+    } else if (current == csharpStringDelimiter) {
```
erwan-leforestier-sonarsource: This closes the C# string on any `"`, but an interpolated string can contain nested string literals inside its `{...}` holes. For valid Razor like `var marker = $"{(true ? "}" : "{")}";`, the `"` before `}` is read as ending the string, so the following `}` is counted as a real brace and closes the `@{ }` block early. The `if`/`else` after it then looks unconditional and S7930 reports a false duplicate id. A test with `$"..."` containing inner quotes would cover this.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/TemplateConditionalScopeTracker.java
```diff
@@ -106,10 +200,62 @@ public void endElement(TagNode node) {
...
+    while (!conditionalBraces.isEmpty() && conditionalBraces.peek().elementDepth() > elementDepth) {
+      recoverConditionalBrace(conditionalBraces.pop());
+    }
+  }
+
+  private void resolvePendingRenderedClosingBrace() {
+    if (pendingRenderedClosingBraceDepth < 0) {
+      return;
+    }
+    if (currentElementDepth() >= pendingRenderedClosingBraceDepth && !conditionalBraces.isEmpty()) {
```
erwan-leforestier-sonarsource, line 218: **False positive: a literal `}` in markup text closes the conditional scope**

When a lone `}` appears in rendered markup text and is followed by a child element (no end tag in between), this guard pops the conditional brace and closes the surrounding `@if`/`else` scope early. Both branches then look unconditional and S7930 fires a false-positive duplicate id.

**Example** (differs only by the `}` in the text):
```razor
@{
    @if (Model.ShowPrimary)
    {
        <div>Total is 100} percent
            <span id="choice">First</span>
        </div>
    }
    else
    {
        <span id="choice">Second</span>
    }
}
```
Result: `S7930 line 10: Duplicate id "choice" found. First occurrence was on line 5.` The same file without the `}` is clean. Reproduced against the compiled scanner.

**Why:** the literal `}` sets `pendingRenderedClosingBraceDepth` at the `<div>` depth. The wrapper's end tag would normally discard it, but the child `<span>` arrives first, so `currentElementDepth() >= pending` holds and the conditional is popped even though the `}` was just text.

**Suggestion:** only resolve a pending brace when it actually balances an opening markup brace, and add a test with a literal `}` in branch markup text followed by a child element.

erwan-leforestier-sonarsource (approved):

PR #808, opened 2026-08-28
SONARHTML-470 Create rule S9379: DOM elements should not use the "autofocus" attribute

guillemsarda:
----
## Summary by Gitar

- **New checks:**
    - Added `NoAutofocusCheck` rule `S9379` to flag `autofocus` attributes on DOM elements
- **Rule definition:**
    - Created rule documentation and JSON metadata for `S9379`
- **Quality profile:**
    - Added rule `S9379` to the Sonar way profile

### DIFF
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoAutofocusCheck.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoAutofocusCheck.java
@@ -31,14 +31,14 @@ public class NoAutofocusCheck extends AbstractPageCheck {
     if (!node.hasProperty("autofocus")) {
       return;
     }
-    if (Helpers.hasAncestorMatching(node, NoAutofocusCheck::isExemptAncestor)) {
+    if (isDialogOrPopover(node) || Helpers.hasAncestorMatching(node, NoAutofocusCheck::isDialogOrPopover)) {
       return;
     }
     createViolation(node, MESSAGE);
   }
 
-  private static boolean isExemptAncestor(TagNode ancestor) {
-    return "dialog".equalsIgnoreCase(ancestor.getNodeName()) || ancestor.hasProperty("popover");
+  private static boolean isDialogOrPopover(TagNode node) {
+    return "dialog".equalsIgnoreCase(node.getNodeName()) || node.hasProperty("popover");
   }
 
 }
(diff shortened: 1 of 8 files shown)


sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoAutofocusCheck.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.sonar.check.Rule;
+import org.sonar.plugins.html.api.Helpers;
+import org.sonar.plugins.html.checks.AbstractPageCheck;
+import org.sonar.plugins.html.node.TagNode;
+
+@Rule(key = "S9379")
+public class NoAutofocusCheck extends AbstractPageCheck {
+
+  private static final String MESSAGE = "The autoFocus prop should not be used, as it can reduce usability and accessibility for users.";
```
nathsou: Non-blocking: this analyzer reports HTML, so `autoFocus prop` is misleading terminology here. Could we use the Jira wording, `Remove this "autofocus" attribute.`?

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoAutofocusCheck.java
```diff
@@ -0,0 +1,44 @@
...
+
+  @Override
+  public void startElement(TagNode node) {
+    if (!node.hasProperty("autofocus")) {
+      return;
+    }
+    if (isDialogOrPopover(node) || Helpers.hasAncestorMatching(node, NoAutofocusCheck::isDialogOrPopover)) {
```
nathsou, line 35: Non-blocking consistency note: this also exempts `autofocus` on the `dialog`/`popover` element itself. HTML permits that behavior, but it differs from SONARHTML-470, the current S9379 wording, and Biome's ancestor-only implementation. Please align those artifacts with whichever scope we want.

nathsou (approved):
Looks good overall. I left two non-blocking consistency and wording suggestions inline.

image

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoAutofocusCheck.java
guillemsarda: Good catch! I actually extended the Jira ticket message with a little reasoning on the why.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoAutofocusCheck.java
guillemsarda: Thank you! I updated the rspec accordingly

PR #815, opened 2026-09-02
SONARHTML-470 Update rule S9379 to cover Vue corner cases

guillemsarda:
----
## Summary by Gitar

- **Rule updates:**
    - Updated `NoAutofocusCheck` to ignore unknown tags, kebab-case tags, and Vue PascalCase components.
    - Added support for ignoring DOM property bindings set to false like `:autofocus="false"`.
- **Helper additions:**
    - Added `startsWithUpperCase` and `isKebabCase` helpers in `Helpers.java` for tag name analysis.

### DIFF
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/Helpers.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/Helpers.java
@@ -137,19 +137,6 @@ public class Helpers {
   }
 
   /**
-   * Returns true when {@code name} looks like a component/custom-element reference rather than a
-   * native HTML tag: PascalCase or kebab-case. Check this ahead of any known-tag whitelist, since
-   * a component can share its name with a native tag case-insensitively
-   * (e.g. Vue's {@code <Input>}, Ant Design's {@code Input} component).
-   *
-   * @param name the tag name to test
-   * @return true if the name is a component reference
-   */
-  public static boolean isComponentReference(String name) {
-    return startsWithUpperCase(name) || isKebabCase(name);
-  }
-
-  /**
    * Returns true if any ancestor of {@code node} satisfies {@code predicate}.
    *
    * @param node the tag node whose ancestors are inspected
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/Helpers.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/Helpers.java
@@ -129,11 +129,6 @@ public class Helpers {
     return false;
   }
 
-  /** Checks if a tag name uses kebab-case; native HTML element names never contain a hyphen, so this always indicates a component. */
-  public static boolean isKebabCase(String name) {
-    return name.indexOf('-') >= 0;
-  }
-
   /**
    * Returns true if any ancestor of {@code node} satisfies {@code predicate}.
    *
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/Helpers.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/Helpers.java
@@ -114,6 +114,27 @@ public class Helpers {
   }
 
   /**
+   * Checks if a tag name uses PascalCase (has an uppercase letter after the first character).
+   * In Vue templates, PascalCase indicates a component reference, not an HTML element.
+   *
+   * @param name the tag name to test
+   * @return true if the name is PascalCase
+   */
+  public static boolean isPascalCase(String name) {
+    for (int i = 1; i < name.length(); i++) {
+      if (Character.isUpperCase(name.charAt(i))) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  /** Checks if a tag name uses kebab-case; native HTML element names never contain a hyphen, so this always indicates a component. */
+  public static boolean isKebabCase(String name) {
+    return name.indexOf('-') >= 0;
+  }
+
+  /**
    * Returns true if any ancestor of {@code node} satisfies {@code predicate}.
    *
    * @param node the tag node whose ancestors are inspected
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/Helpers.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/Helpers.java
@@ -114,19 +114,39 @@ public class Helpers {
   }
 
   /**
-   * Checks if a tag name uses PascalCase (has an uppercase letter after the first character).
-   * In Vue templates, PascalCase indicates a component reference, not an HTML element.
+   * Returns true when {@code name} starts with an uppercase letter, the convention Vue (and other
+   * frameworks) use to write a PascalCase component reference in a template.
    *
    * @param name the tag name to test
-   * @return true if the name is PascalCase
+   * @return true if the first character is uppercase
    */
-  public static boolean isPascalCase(String name) {
-    for (int i = 1; i < name.length(); i++) {
-      if (Character.isUpperCase(name.charAt(i))) {
-        return true;
-      }
-    }
-    return false;
+  public static boolean startsWithUpperCase(String name) {
+    return !name.isEmpty() && Character.isUpperCase(name.charAt(0));
+  }
+
+  /**
+   * Returns true when {@code name} contains a hyphen. No native HTML tag name has one, so a
+   * hyphen marks a custom element or component reference (Vue kebab-case, Angular/Web Components
+   * selectors).
+   *
+   * @param name the tag name to test
+   * @return true if the name contains a hyphen
+   */
+  public static boolean isKebabCase(String name) {
+    return name.indexOf('-') >= 0;
+  }
+
+  /**
+   * Returns true when {@code name} looks like a component/custom-element reference rather than a
+   * native HTML tag: PascalCase or kebab-case. Check this ahead of any known-tag whitelist, since
+   * a component can share its name with a native tag case-insensitively
+   * (e.g. Vue's {@code <Input>}, Ant Design's {@code Input} component).
+   *
+   * @param name the tag name to test
+   * @return true if the name is a component reference
+   */
+  public static boolean isComponentReference(String name) {
+    return startsWithUpperCase(name) || isKebabCase(name);
   }
 
   /**
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoAutofocusCheck.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoAutofocusCheck.java
@@ -36,8 +36,12 @@ public class NoAutofocusCheck extends AbstractPageCheck {
     if (autofocusProperty == null) {
       return;
     }
-    // Components/custom elements use `autofocus` as a prop, not the DOM attribute; checked before the tag whitelist since names can collide case-insensitively, e.g. Vue's <Input>.
-    if (Helpers.isComponentReference(node.getNodeName()) || !hasKnownHTMLTag(node)) {
+    // Kebab-case is always a custom element (no native tag has a hyphen); PascalCase only means a
+    // component in Vue, since HTML tag names are otherwise case-insensitive, e.g. plain <BUTTON>.
+    String nodeName = node.getNodeName();
+    boolean componentReference = Helpers.isKebabCase(nodeName)
+      || (Helpers.isVueFile(getHtmlSourceCode()) && Helpers.startsWithUpperCase(nodeName));
+    if (componentReference || !hasKnownHTMLTag(node)) {
       return;
     }
     // A DOM-property binding (`:x`, `v-bind:x`, `[x]`) bound to literal false never sets the property, unlike a static "false" string.
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoAutofocusCheck.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoAutofocusCheck.java
@@ -17,9 +17,12 @@
 package org.sonar.plugins.html.checks.accessibility;
 
 import java.util.Arrays;
+import java.util.List;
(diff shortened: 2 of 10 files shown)


sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/Helpers.java
```diff
@@ -113,6 +113,22 @@ public static boolean isVueFile(HtmlSourceCode code) {
...
   }
 
+  /**
+   * Checks if a tag name uses PascalCase (has an uppercase letter after the first character).
+   * In Vue templates, PascalCase indicates a component reference, not an HTML element.
+   *
+   * @param name the tag name to test
+   * @return true if the name is PascalCase
+   */
+  public static boolean isPascalCase(String name) {
```
martin-strecker-sonarsource: `isPascalCase` skips the first character (loop starts at `i=1`), so `Modal`/`Input`/`Header` → `false` and `customInput` (camelCase) → `true` - both wrong.

Verified against Vue's own compiler: it classifies purely via `isUpperCase(tag.charCodeAt(0))`, nothing more.
https://github.com/vuejs/core/blob/d105e19e914bca6e1140bfec074e055ed0679e21/packages/compiler-core/src/parser.ts#L782-L798
https://github.com/vuejs/core/blob/d105e19e914bca6e1140bfec074e055ed0679e21/packages/compiler-core/src/parser.ts#L832-L834

Suggest `Character.isUpperCase(name.charAt(0))`. Naming nit (non-blocking): `startsWithUpperCase` fits better than `isPascalCase`.

Also: its use in `NoAutofocusCheck` was already abandoned within this PR (added in d0e63a85, replaced by `hasKnownHTMLTag` in 9ed198d6) - only `UnsupportedTagsInHtml5Check` needs it now. See the `hasKnownHTMLTag` comment for the combined fix.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoAutofocusCheck.java
```diff
@@ -20,16 +20,28 @@
...
 
   @Override
   public void startElement(TagNode node) {
-    if (!node.hasProperty("autofocus")) {
+    Attribute autofocusProperty = node.getProperty("autofocus");
+    if (autofocusProperty == null) {
+      return;
+    }
+    // Components and custom elements are skipped: `autofocus` there is a prop, not the DOM attribute.
+    if (!hasKnownHTMLTag(node)) {
```
martin-strecker-sonarsource: `hasKnownHTMLTag` is case-insensitive, so a component literally named `Input`/`Table`/`Dialog` (e.g. ant-design's `Input`) written as `<Input autofocus />` isn't exempted - falls through to `createViolation`, the exact FP this PR targets. `VueComponents.vue` doesn't cover this - please add a case before merging.

Checked Vue/Angular source rather than assuming:
- Vue classifies by `isUpperCase(tag.charCodeAt(0))` alone, *before* consulting any tag whitelist: https://github.com/vuejs/core/blob/d105e19e914bca6e1140bfec074e055ed0679e21/packages/compiler-core/src/parser.ts#L782-L798
- Angular has no case-based detection at all - pure selector/schema matching, kebab-case is convention only: https://github.com/angular/angular/blob/81c3f3a6aab30a0e9ca3a8b511100afea4d07003/packages/compiler-cli/src/ngtsc/typecheck/src/dom.ts#L41-L79 and https://github.com/angular/angular/blob/81c3f3a6aab30a0e9ca3a8b511100afea4d07003/packages/compiler/src/schema/dom_element_schema_registry.ts#L416-L430

Recommendation:
1. Fix `isPascalCase` → `Character.isUpperCase(name.charAt(0))`
2. Restore `isKebabCase` (`name.indexOf('-') >= 0` - no native tag has a hyphen, so this alone is sufficient)
3. Give both priority over the whitelist: `if (startsWithUpperCase(nodeName) || isKebabCase(nodeName) || !hasKnownHTMLTag(node)) return;`
4. One `isComponentReference` combinator rather than per-framework helpers - Angular needs nothing beyond the shared kebab check.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/UnsupportedTagsInHtml5Check.java
```diff
@@ -72,24 +72,11 @@ private boolean isUnsupportedTag(TagNode node) {
 
     // In Vue files, PascalCase tags are components, not HTML elements
     // e.g., <BLink> is a Vue Bootstrap component, not the deprecated <blink> tag
-    if (isVueFile && isPascalCase(nodeName)) {
+    if (isVueFile && Helpers.isPascalCase(nodeName)) {
       return false;
     }
 
     return UNSUPPORTED_TAGS.contains(nodeName.toUpperCase(Locale.ENGLISH));
```
martin-strecker-sonarsource: Not a bug - `Locale.ENGLISH` correctly avoids the Turkish-I issue (several entries here contain 'i': LISTING, BLINK, DIR, ISINDEX, NEXTID, PLAINTEXT, STRIKE). Minor consistency suggestion: `HtmlConstants.hasKnownHTMLTag` already uses locale-independent `equalsIgnoreCase` instead of transform-then-contains.

1. `UNSUPPORTED_TAGS.stream().anyMatch(nodeName::equalsIgnoreCase)` - one-line swap.
2. `TreeSet<>(String.CASE_INSENSITIVE_ORDER)` - note `TreeSet` has no `of(...)` factory, so this needs a constructor+addAll, not a drop-in replacement.

Option 1 is simpler here.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoAutofocusCheck.java
```diff
@@ -20,16 +20,28 @@
...
+
 @Rule(key = "S9379")
 public class NoAutofocusCheck extends AbstractPageCheck {
 
   private static final String MESSAGE = "Remove this \"autofocus\" attribute, as it can reduce usability and accessibility for users.";
 
   @Override
   public void startElement(TagNode node) {
-    if (!node.hasProperty("autofocus")) {
+    Attribute autofocusProperty = node.getProperty("autofocus");
```
martin-strecker-sonarsource, line 44: Style suggestion: only the first guard (`autofocusProperty == null`) is a genuine extract-and-bail; the other 3 are really one composite decision. `NoAriaHiddenOnFocusableCheck` in this same package already does the split you'd expect:
```java
if (!hasKnownHTMLTag(node)) { return; }
if (isFocusableElement(node) && !isHiddenInput(node) && hasAriaHidden(node)) { createViolation(...); }
```
Suggest the same shape here:
```java
if (autofocusProperty == null) { return; }
if (hasKnownHTMLTag(node) && !isDomPropertyBoundToFalse(autofocusProperty) && !isDialogOrPopover(node)
    && !Helpers.hasAncestorMatching(node, NoAutofocusCheck::isDialogOrPopover)) {
  createViolation(node, MESSAGE);
}
```
Trade-off: the per-guard comments would need to move above the block or get dropped.

martin-strecker-sonarsource (changes requested):
Deep-dived S9379 component-vs-native-tag detection here. Found a real correctness bug (hasKnownHTMLTag/isPascalCase collision reintroduces the exact false-positive class this PR is meant to fix, for any component named after a real tag e.g. Input/Table/Dialog) plus a couple of smaller suggestions. Details inline.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/Helpers.java
guillemsarda: Fixed as suggested: renamed to `startsWithUpperCase`, reimplemented as `Character.isUpperCase(name.charAt(0))` (matches Vue's own classification, as you linked). `UnsupportedTagsInHtml5Check` calls it directly; `NoAutofocusCheck` goes through the new `isComponentReference` combinator described in the reply below.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoAutofocusCheck.java
guillemsarda: Applied all four points: `startsWithUpperCase` now checks only the first character, `isKebabCase` is back (`name.indexOf('-') >= 0`), both combined into `isComponentReference`, checked with priority over `hasKnownHTMLTag`: `if (Helpers.isComponentReference(node.getNodeName()) || !hasKnownHTMLTag(node)) return;`. Added a `<Input autofocus />` case to `VueComponents.vue` covering the case-collision scenario and updated the expected line in the test.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/UnsupportedTagsInHtml5Check.java
guillemsarda: Went with option 1: `UNSUPPORTED_TAGS.stream().anyMatch(nodeName::equalsIgnoreCase)`, and dropped the now-unused `Locale` import.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoAutofocusCheck.java
guillemsarda: Kept the same extract-and-bail + composite shape you pointed to, but left the tag-whitelist check as its own early-return guard rather than folding it into the positive composite: it now needs to short-circuit together with the new `isComponentReference` check (`isComponentReference(...) || !hasKnownHTMLTag(...)`), which reads awkwardly if inverted inside the createViolation condition. That also let me keep the per-guard comments instead of dropping them, per the trade-off you flagged.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoAutofocusCheck.java
```diff
@@ -20,22 +20,40 @@
...
+    if (autofocusProperty == null) {
       return;
     }
-    if (isDialogOrPopover(node) || Helpers.hasAncestorMatching(node, NoAutofocusCheck::isDialogOrPopover)) {
+    // Components/custom elements use `autofocus` as a prop, not the DOM attribute; checked before the tag whitelist since names can collide case-insensitively, e.g. Vue's <Input>.
+    if (Helpers.isComponentReference(node.getNodeName()) || !hasKnownHTMLTag(node)) {
       return;
     }
```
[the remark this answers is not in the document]
guillemsarda: Good catch. Applied your Fix 2, generalized slightly: kept the kebab-case exemption ungated (no native tag ever has a hyphen, so it's safe in any file type — covers Angular/Web Components too), and gated only the PascalCase branch on `isVueFile`, since HTML tag case is otherwise meaningless — `<BUTTON>`/`<INPUT>` in plain HTML are just the native tags typed in caps. Removed `isComponentReference` (no longer had a caller once the two branches split apart) along with its test. Added `<BUTTON autofocus>`/`<INPUT AUTOFOCUS />` cases to `CustomElementInHtmlFile.html` asserting they're still flagged.

sonar-html-plugin/src/main/resources/org/sonar/l10n/web/rules/Web/S9379.html
```diff
@@ -42,6 +42,13 @@ <h4>Compliant solution</h4>
   &lt;input autofocus /&gt;
 &lt;/div&gt;
 </pre>
+<p>An <code>autofocus</code> attribute on a tag that is not a recognized native HTML element (a Vue or Angular component, or a custom element, in
```
martin-strecker-sonarsource, line 45: This paragraph is stale — it describes the behavior from before the Gitar-feedback fix. It says PascalCase (like kebab-case) exempts a tag "regardless" of context, but the final logic only exempts PascalCase inside `.vue` files (`NoAutofocusCheck.java:42-43`); outside Vue, a PascalCase name that collides with a real tag (`<BUTTON>`, `<Input>`) is still flagged, per `CustomElementInHtmlFile.html`. Attributing PascalCase detection to "Angular" is also off — Angular selectors are conventionally kebab-case, not PascalCase.

Suggest rewording, e.g.:
"Kebab-case always indicates a component/custom element (no native tag has a hyphen). PascalCase indicates a component only in Vue templates — outside Vue, HTML tag names are case-insensitive, so `<BUTTON autofocus>` is still the native tag and is flagged."

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/NoAutofocusCheck.java
```diff
@@ -20,22 +20,44 @@
...
+    Attribute autofocusProperty = node.getProperty("autofocus");
+    if (autofocusProperty == null) {
       return;
     }
-    if (isDialogOrPopover(node) || Helpers.hasAncestorMatching(node, NoAutofocusCheck::isDialogOrPopover)) {
+    // Kebab-case is always a custom element (no native tag has a hyphen); PascalCase only means a
+    // component in Vue, since HTML tag names are otherwise case-insensitive, e.g. plain <BUTTON>.
+    String nodeName = node.getNodeName();
+    boolean componentReference = Helpers.isKebabCase(nodeName)
+      || (Helpers.isVueFile(getHtmlSourceCode()) && Helpers.startsWithUpperCase(nodeName));
```
martin-strecker-sonarsource: `Helpers.isVueFile(getHtmlSourceCode())` is now recomputed on every `startElement` call (once per tag) instead of being cached. `UnsupportedTagsInHtml5Check` — touched in this same PR — does this correctly: computes it once in `startDocument()` into an `isVueFile` field. `NoAutofocusCheck` actually had that same field/`startDocument` override earlier in this PR's history but it got dropped along the way.

Not a correctness issue (the check itself is cheap), but worth restoring the field for consistency with the sibling check in this PR.

martin-strecker-sonarsource (approved):
Minor NITs that should be addressed but don't need a further review

PR #819, opened 2026-09-04
SONARHTML-471 Implement rule S9380: The "scope" attribute should only be used on "th" elements- #7876

guillemsarda:
Part of 

## Disclaimer
- This PR contains the helpers pending to be merged in this other [PR](https://github.com/SonarSource/sonar-html/pull/815).

----
## Summary by Gitar

- **New accessibility rule:**
    - Implemented rule `S9380` to ensure the `scope` attribute is only used on `th` elements.

### DIFF
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/ScopeAttributeOnlyOnThCheck.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/ScopeAttributeOnlyOnThCheck.java
@@ -41,7 +41,8 @@ public class ScopeAttributeOnlyOnThCheck extends AbstractPageCheck implements Em
 
   @Override
   public void startElement(TagNode node) {
-    if (!hasScopeAttribute(node) || "th".equalsIgnoreCase(node.getNodeName()) || isTemplateTag(node)) {
+    // Vue 2.0-2.4 scoped slots use a bare "scope" attribute on <template>, unrelated to table headers.
+    if (!hasScopeAttribute(node) || "th".equalsIgnoreCase(node.getNodeName()) || Helpers.isTemplateLikeTag(node)) {
       return;
     }
     // A custom component or unknown tag: "scope" may be an arbitrary prop, unrelated to table headers.
@@ -54,11 +55,6 @@ public class ScopeAttributeOnlyOnThCheck extends AbstractPageCheck implements Em
   private static boolean hasScopeAttribute(TagNode node) {
     return node.getAttributes().stream()
       .anyMatch(a -> SCOPE_ATTRIBUTE_NAMES.contains(a.getName().toLowerCase(Locale.ROOT)));
-  }
-
-  private static boolean isTemplateTag(TagNode node) {
-    // Vue 2.0-2.4 scoped slots use a bare "scope" attribute on <template>, unrelated to table headers.
-    return "template".equalsIgnoreCase(node.getNodeName());
   }
 
   private boolean isComponentReference(TagNode node) {
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/ScopeAttributeOnlyOnThCheck.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/ScopeAttributeOnlyOnThCheck.java
@@ -16,6 +16,10 @@
  */
 package org.sonar.plugins.html.checks.accessibility;
 
+import java.util.Locale;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.sonar.check.Rule;
 import org.sonar.plugins.html.api.Helpers;
 import org.sonar.plugins.html.api.HtmlConstants;
@@ -28,9 +32,16 @@ public class ScopeAttributeOnlyOnThCheck extends AbstractPageCheck implements Em
 
   private static final String MESSAGE = "Move this \"scope\" attribute to a \"th\" element, or remove it.";
 
+  // Spellings of a "scope" attribute; excludes Vue's ":[scope]" dynamic argument, whose target is only known at runtime.
+  private static final Set<String> SCOPE_ATTRIBUTE_NAMES = Stream.concat(
+      Stream.of("scope", "[attr.scope]", "attr.scope"),
+      TagNode.domPropertyBindingNames("scope").stream())
+    .map(name -> name.toLowerCase(Locale.ROOT))
+    .collect(Collectors.toUnmodifiableSet());
+
   @Override
   public void startElement(TagNode node) {
-    if (!node.hasProperty("scope") || "th".equalsIgnoreCase(node.getNodeName()) || isTemplateTag(node)) {
+    if (!hasScopeAttribute(node) || "th".equalsIgnoreCase(node.getNodeName()) || isTemplateTag(node)) {
       return;
     }
     // A custom component or unknown tag: "scope" may be an arbitrary prop, unrelated to table headers.
@@ -38,6 +49,11 @@ public class ScopeAttributeOnlyOnThCheck extends AbstractPageCheck implements Em
       return;
     }
     createViolation(node, MESSAGE);
+  }
+
+  private static boolean hasScopeAttribute(TagNode node) {
+    return node.getAttributes().stream()
+      .anyMatch(a -> SCOPE_ATTRIBUTE_NAMES.contains(a.getName().toLowerCase(Locale.ROOT)));
   }
 
   private static boolean isTemplateTag(TagNode node) {
--- a/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/ScopeAttributeOnlyOnThCheck.java
+++ b/sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/ScopeAttributeOnlyOnThCheck.java
@@ -1,0 +1,62 @@
+/*
+ * SonarQube HTML
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * You can redistribute and/or modify this program under the terms of
+ * the Sonar Source-Available License Version 1, as published by SonarSource Sàrl.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+package org.sonar.plugins.html.checks.accessibility;
+
+import org.sonar.check.Rule;
+import org.sonar.plugins.html.api.Helpers;
+import org.sonar.plugins.html.api.HtmlConstants;
+import org.sonar.plugins.html.checks.AbstractPageCheck;
+import org.sonar.plugins.html.checks.EmbeddedHtmlCheck;
+import org.sonar.plugins.html.node.TagNode;
+
+@Rule(key = "S9380")
+public class ScopeAttributeOnlyOnThCheck extends AbstractPageCheck implements EmbeddedHtmlCheck {
+
+  private static final String MESSAGE = "Move this \"scope\" attribute to a \"th\" element, or remove it.";
+
+  @Override
+  public void startElement(TagNode node) {
+    if (!node.hasProperty("scope") || "th".equalsIgnoreCase(node.getNodeName()) || isTemplateTag(node)) {
+      return;
+    }
+    // A custom component or unknown tag: "scope" may be an arbitrary prop, unrelated to table headers.
+    if (isComponentReference(node) || !HtmlConstants.hasKnownHTMLTag(node)) {
+      return;
+    }
+    createViolation(node, MESSAGE);
+  }
+
+  private static boolean isTemplateTag(TagNode node) {
+    // Vue 2.0-2.4 scoped slots use a bare "scope" attribute on <template>, unrelated to table headers.
+    return "template".equalsIgnoreCase(node.getNodeName());
+  }
+
+  private boolean isComponentReference(TagNode node) {
+    String nodeName = node.getNodeName();
+    // Kebab-case is always a custom element; PascalCase only means a component in Vue files.
+    return isKebabCase(nodeName) || (Helpers.isVueFile(getHtmlSourceCode()) && startsWithUpperCase(nodeName));
+  }
+
+  private static boolean isKebabCase(String name) {
+    return name.indexOf('-') >= 0;
+  }
+
+  private static boolean startsWithUpperCase(String name) {
+    return !name.isEmpty() && Character.isUpperCase(name.charAt(0));
+  }
+
+}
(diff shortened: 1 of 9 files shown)


sonar-html-plugin/src/main/resources/org/sonar/l10n/web/rules/Web/S9380.html
```diff
@@ -0,0 +1,51 @@
...
+<h3>Code examples</h3>
+<h4>Noncompliant code example</h4>
+<pre data-diff-id="1" data-diff-type="noncompliant">
+&lt;table&gt;
+  &lt;tr&gt;
+    &lt;td scope="row"&gt;Total&lt;/td&gt;
+    &lt;td&gt;42&lt;/td&gt;
+  &lt;/tr&gt;
+&lt;/table&gt;
+</pre>
```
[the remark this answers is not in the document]
guillemsarda: We already knew this and we accept it as a valid overlap since it is partial for S1827 and the motivation behind the rules is different.

martin-strecker-sonarsource:
We looked at the violations in `its/ruling/src/test/resources/expected/Web-S9380.json` — 348 flagged locations, but they all cluster into just 7 distinct line-number/file-family groups (`HTMLCollection*`, `HTMLTableCellElement*`, `HTMLTableElement*`, `HTMLTableRowElement*`, `HTMLTableSectionElement*`, `table*`, in both `.html` and `.xhtml` variants), all within a single project: the vendored NIST DOM Level 2 HTML conformance suite under `external_webkit-jb-mr1`.

We sampled one file from each of the 7 groups directly against the source (`SonarCommunity/web-test-sources`) and every one hits the same boilerplate fixture line, e.g.:
```html
<TD AXIS="center" ID="Table-3" ABBR="maiden2" ... ROWSPAN="1" SCOPE="col" HEADERS="header-2" ...>EMP0001</TD>
```
All 348 are genuine `scope` attributes on `<td>` elements — true positives, not false positives. The huge count is one legacy fixture duplicated ~150+ times across numbered test variants (01, 02, 03...), not a rule that's over-firing.

One thing worth flagging: these files declare `<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN">`, and HTML 4.01 did permit `scope` on `td`. We considered whether the rule should relax for non-HTML5 doctypes, and don't think it should:

- A DOCTYPE only selects the browser's CSS quirks/limited-quirks/no-quirks rendering mode — box model, font inheritance, cell sizing. It has no effect on the algorithm browsers/assistive technology use to associate header cells via `scope`/`headers`/`id`; that runs identically in every evergreen browser regardless of declared doctype. So a screen-reader user visiting one of these pages today gets no benefit from `scope="col"` on that `td`, HTML4 doctype or not — the accessibility problem is just as real.
- It'd also be inconsistent with `UnsupportedTagsInHtml5Check` in this same codebase, which already flags deprecated tags unconditionally with no doctype gating, despite those tags being valid under older doctypes too. The existing convention here is "reflect modern web-platform/AT behavior, not the page's self-declared doctype."

Given that, might be worth adding a line to `S9380.html` preempting this exact question for future readers, e.g.:

> "This applies even to documents declaring an HTML 4.01 doctype: the doctype only switches the browser's rendering mode, not the algorithm assistive technology uses to associate header cells, so `scope` on a non-`th` element provides no benefit today regardless of the declared doctype."

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/ScopeAttributeOnlyOnThCheck.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+    // A custom component or unknown tag: "scope" may be an arbitrary prop, unrelated to table headers.
+    if (isComponentReference(node) || !HtmlConstants.hasKnownHTMLTag(node)) {
+      return;
+    }
+    createViolation(node, MESSAGE);
+  }
+
+  private static boolean hasScopeAttribute(TagNode node) {
+    return node.getAttributes().stream()
```
martin-strecker-sonarsource: `hasScopeAttribute` flags any attribute literally named `scope`/`:scope`/`v-bind:scope`/`[scope]`/`[attr.scope]`/`attr.scope` regardless of the bound value. Per Vue's own `patchAttr` (`vuejs/core`, `packages/runtime-dom/src/modules/attrs.ts`):
```ts
if (value == null || (isBoolean && !includeBooleanAttr(value))) {
  el.removeAttribute(key)
} else {
  el.setAttribute(key, isBoolean ? '' : isSymbol(value) ? String(value) : value)
}
```
`value == null` (covers both `null` and `undefined`) removes the attribute for *any* attribute, boolean or not — not just recognized boolean attrs. So `<div :scope="null">` / `<div :scope="undefined">` never actually renders a `scope` attribute at runtime, yet this check would still flag it. This is the same FP class fixed for `autofocus` in #815 via `isDomPropertyBoundToFalse`. Suggest an equivalent guard here for literal `"null"`/`"undefined"` values, or confirm this is an accepted gap for the first cut.

Note `:scope="false"` is different — Vue sets the literal string `"false"` for non-boolean attributes, so flagging that one is correct as-is.

sonar-html-plugin/src/main/resources/org/sonar/l10n/web/rules/Web/S9380.html
```diff
@@ -0,0 +1,51 @@
...
+<h4>Compliant solution</h4>
+<pre data-diff-id="1" data-diff-type="compliant">
+&lt;table&gt;
+  &lt;tr&gt;
+    &lt;th scope="row"&gt;Total&lt;/th&gt;
+    &lt;td&gt;42&lt;/td&gt;
+  &lt;/tr&gt;
+&lt;/table&gt;
+</pre>
+<p>A <code>scope</code> attribute on a tag that is not a recognized native HTML element (a Vue or Angular component, or a custom element, in
```
martin-strecker-sonarsource: "A `scope` attribute on a tag that is not a recognized native HTML element (a Vue or Angular component, or a custom element, in PascalCase or kebab-case) is not flagged..." — the code (`isComponentReference`) only exempts PascalCase inside `.vue` files; outside Vue a PascalCase name colliding with a real tag (e.g. `<Caption scope="row">`) is still flagged, per this PR's own `invalid.html:11`. Looks copy-pasted from S9379's pre-fix wording, including the "Angular"-for-PascalCase attribution (Angular selectors are conventionally kebab-case). Same fix as suggested on #815 applies here: clarify kebab-case is universal, PascalCase is Vue-only.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/ScopeAttributeOnlyOnThCheck.java
```diff
@@ -0,0 +1,74 @@
...
+      .anyMatch(a -> SCOPE_ATTRIBUTE_NAMES.contains(a.getName().toLowerCase(Locale.ROOT)));
+  }
+
+  private boolean isComponentReference(TagNode node) {
+    String nodeName = node.getNodeName();
+    // Kebab-case is always a custom element; PascalCase only means a component in Vue files.
+    return isKebabCase(nodeName) || (Helpers.isVueFile(getHtmlSourceCode()) && startsWithUpperCase(nodeName));
+  }
+
+  private static boolean isKebabCase(String name) {
```
martin-strecker-sonarsource: `isKebabCase`/`startsWithUpperCase` duplicate `Helpers.isKebabCase`/`Helpers.startsWithUpperCase` from #815 — already disclosed in the PR description as pending that merge. Flagging as a concrete follow-up: once #815 lands, rebase this PR to drop the private duplicates and call `Helpers.*` directly.

sonar-html-plugin/src/test/resources/checks/ScopeAttributeOnlyOnThCheck/invalid.html
```diff
@@ -0,0 +1,11 @@
+<td scope="row">Total</td>
+<td scope="col">Total</td>
+<div scope="row"></div>
+<span :scope="dynamicScope"></span>
+<div>
+  <td scope="rowgroup">Total</td>
+</div>
+<TD SCOPE="colgroup">Total</TD>
+<Th scope="row">Total</Th>
```
martin-strecker-sonarsource: This line (`<Th scope="row">Total</Th>`) is byte-identical to `valid.html:9`, and `ScopeAttributeOnlyOnThCheckTest.java:47-53` confirms it's deliberately excluded from the expected violations (list is 1,2,3,4,6,8,11 — line 9 skipped). It's a non-violating example living in `invalid.html`, duplicating a case already covered in `valid.html`. Please remove it from `invalid.html`.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/ScopeAttributeOnlyOnThCheck.java
```diff
@@ -0,0 +1,74 @@
...
+import org.sonar.plugins.html.node.TagNode;
+
+@Rule(key = "S9380")
+public class ScopeAttributeOnlyOnThCheck extends AbstractPageCheck implements EmbeddedHtmlCheck {
+
+  private static final String MESSAGE = "Move this \"scope\" attribute to a \"th\" element, or remove it.";
+
+  // Spellings of a "scope" attribute; excludes Vue's ":[scope]" dynamic argument, whose target is only known at runtime.
+  private static final Set<String> SCOPE_ATTRIBUTE_NAMES = Stream.concat(
+      Stream.of("scope", "[attr.scope]", "attr.scope"),
```
martin-strecker-sonarsource: Suggestion: `SCOPE_ATTRIBUTE_NAMES` hand-lists `"scope", "[attr.scope]", "attr.scope"` and concatenates `TagNode.domPropertyBindingNames("scope")` — reimplementing most of what `TagNode.getProperty()` already matches internally (everything except `:[x]`, excluded on purpose per this check's own comment). Suggest a new `TagNode` helper built *on top of* `domPropertyBindingNames` (not folded into it — `NoAutofocusCheck.isDomPropertyBoundToFalse` depends on that method's narrow 3-spelling contract):
```java
/**
 * Every spelling that can set {@code propertyName}, except Vue's dynamic-argument form
 * ({@code :[x]}), whose target name is only known at runtime:
 * <ul>
 *   <li>{@code x} — a plain, static HTML attribute.</li>
 *   <li>{@code [x]} — Angular property binding: sets the DOM property directly.</li>
 *   <li>{@code v-bind:x}, {@code :x} — Vue's {@code v-bind} directive (long form and shorthand).</li>
 *   <li>{@code [attr.x]} — Angular attribute binding: forces setting the HTML attribute itself.</li>
 *   <li>{@code attr.x} (no brackets) — Angular's static form of the same prefix: per SONARHTML-158,
 *       the right-hand side is used as a literal string, not evaluated as an expression.</li>
 * </ul>
 */
public static List<String> allAttributeSpellings(String propertyName) {
  return Stream.concat(
    Stream.of(propertyName, "[attr." + propertyName + "]", "attr." + propertyName),
    domPropertyBindingNames(propertyName).stream())
    .toList();
}
```
`SCOPE_ATTRIBUTE_NAMES` then becomes `TagNode.allAttributeSpellings("scope").stream().map(toLowerCase)...`, and `getProperty()` could reuse the same helper for its first four match arms, leaving only `vueSquaredShorthandProperty` as its own explicitly-commented case. One place ends up owning every spelling, with terminology that matches what each syntax actually does — `getProperty()`'s current javadoc just says "takes into account the property binding mechanism of angular and vue.js" without naming any of it.

martin-strecker-sonarsource (changes requested):
Found a likely false positive plus a few NITs. Requesting changes mainly for the null/undefined binding gap.

sonar-html-plugin/src/main/resources/org/sonar/l10n/web/rules/Web/S9380.html
guillemsarda: This was just a mock temp file. I've updated it anyway with the info in the [rspec PR](https://github.com/SonarSource/rspec/pull/8070).

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/ScopeAttributeOnlyOnThCheck.java
guillemsarda: It makes sense. I think it's worth tackling this in a [different task](https://sonarsource.atlassian.net/browse/SONARHTML-486) since it would be interesting to reuse the helper also in the recently merged rule S9379. But using this helper in this rule would cause a behavior change, polluting this PR with unrelated changes.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/accessibility/ScopeAttributeOnlyOnThCheck.java
```diff
@@ -0,0 +1,82 @@
...
+
+  // Spellings of a "scope" attribute; excludes Vue's ":[scope]" dynamic argument, whose target is only known at runtime.
+  private static final Set<String> SCOPE_ATTRIBUTE_NAMES = Stream.concat(
+      Stream.of(SCOPE, "[attr.scope]", "attr.scope"),
+      TagNode.domPropertyBindingNames(SCOPE).stream())
+    .map(name -> name.toLowerCase(Locale.ROOT))
+    .collect(Collectors.toUnmodifiableSet());
+
+  @Override
+  public void startElement(TagNode node) {
```
martin-strecker-sonarsource, line 52: nit: `SCOPE_ATTRIBUTE_NAMES` includes the Angular attribute-binding spellings `"[attr.scope]"` and `"attr.scope"`, but neither `invalid.html` nor `valid.html` has a test case exercising either spelling - so that part of the matching logic is currently untested. Suggest adding one case per spelling (e.g. `<td [attr.scope]="'row'">`/`<td attr.scope="row">` to `invalid.html`) to cover them.

martin-strecker-sonarsource (approved):
Approving - nit: `SCOPE_ATTRIBUTE_NAMES` includes the Angular attribute-binding spellings `"[attr.scope]"` and `"attr.scope"`, but neither `invalid.html` nor `valid.html` exercises either spelling, so that part of the matching logic is currently untested. Non-blocking.

sonar-html-plugin/src/main/resources/org/sonar/l10n/web/rules/Web/S9380.html
```diff
@@ -0,0 +1,47 @@
...
+get no benefit from it, and the markup becomes invalid.</p>
+<p>The <code>td</code> element is a special case: HTML4 allowed <code>scope</code> on a data cell, but HTML5 deprecated that usage, keeping
+<code>scope</code> valid only on <code>th</code>. Code that still carries <code>scope</code> on a <code>td</code> is most likely leftover from that
+older practice.</p>
+<h2>How to fix it</h2>
+<p>Move the <code>scope</code> attribute to the <code>th</code> element that acts as the header for the row, column, row group, or column group, or
+remove the attribute if the element is not a table header.</p>
+<p>Custom elements and components (kebab-case tags, PascalCase Vue components) are not flagged, since <code>scope</code> may be an arbitrary prop
+unrelated to the native HTML attribute. The bare <code>scope</code> attribute on Vue’s <code>&lt;template&gt;</code> (2.x scoped slots) is also
+excluded, as it is unrelated to table headers.</p>
```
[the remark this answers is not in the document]
guillemsarda: Out of scope, the rule description is judged in the rspec PR not in SONARHTML

PR #825, opened 2026-09-08
SONARHTML-487 Fix duplicate-ID false positives in mutually exclusive template branches

erwan-leforestier-sonarsource:
## Summary

S7930 reports duplicate HTML `id` attributes, which can cause conflicts when the elements are rendered on the same page.

This change prevents a false positive when identical IDs are nested in mutually exclusive Angular or Vue template branches. It still reports duplicate IDs whose elements can be rendered together.

## False positive fixed

Only one of these Angular branches can render, so `id="search"` is not duplicated in the resulting page:

```html
<div *ngIf="compactMode">
  <input id="search" />
</div>
<div *ngIf="!compactMode">
  <input id="search" />
</div>
```

Before this change, S7930 raised an issue for the two nested `search` IDs. It no longer does.


sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/TemplateConditionalScopeTracker.java
```diff
@@ -237,7 +248,143 @@ private int currentElementDepth() {
...
+    if (firstScope.hasAttribute(ANGULAR_SWITCH_DEFAULT_ATTRIBUTE)) {
+      return !secondScope.hasAttribute(ANGULAR_SWITCH_DEFAULT_ATTRIBUTE) && secondScope.hasAttribute(ANGULAR_SWITCH_CASE_ATTRIBUTE);
+    }
+    if (secondScope.hasAttribute(ANGULAR_SWITCH_DEFAULT_ATTRIBUTE)) {
+      return firstScope.hasAttribute(ANGULAR_SWITCH_CASE_ATTRIBUTE);
+    }
+    String firstCase = firstScope.getAttribute(ANGULAR_SWITCH_CASE_ATTRIBUTE);
+    String secondCase = secondScope.getAttribute(ANGULAR_SWITCH_CASE_ATTRIBUTE);
+    return firstCase != null && secondCase != null && areDistinctLiteralValues(firstCase, secondCase);
+  }
```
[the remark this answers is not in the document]
erwan-leforestier-sonarsource: Not applying this change. Angular NgSwitch renders every matching case, and distinct expression text does not prove distinct runtime values (for example, two properties or enum aliases can resolve to the same value). The checker therefore treats only statically distinct literals as mutually exclusive; this preserves reporting for possible duplicate IDs. The switch-case regression test verifies the proven-literal case.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/coding/NoDuplicateIDCheck.java
```diff
@@ -80,28 +85,83 @@ private void handleIdAttribute(TagNode node) {
...
+    List<TagNode> attributeScopes = conditionalScope.conditionalAttributeScopes(node);
+    if (!attributeScopes.isEmpty()) {
+      if (!reportDuplicateAgainstUnconditionalId(node, idValue)) {
+        registerConditionalId(node, idValue, attributeScopes);
+      }
+    } else if (conditionalScope.isConditionalAttributeHost(node)) {
       reportDuplicateAgainstUnconditionalId(node, idValue);
     } else {
       registerUnconditionalId(node, idValue);
     }
```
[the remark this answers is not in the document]
erwan-leforestier-sonarsource: Not applying this change. This source-order treatment of IDs on v-for/*ngFor hosts predates this PR: before the change, isInConditional classified loop hosts as conditional and only checked them against an already recorded unconditional ID. Commit 2194a3ca deliberately kept loops out of conditional-branch paths so a repeater cannot give descendants branch-exclusivity semantics. Tracking the host in conditionalIds would change that established S7930 behavior; this PR only changes descendant IDs under conditional branches.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/coding/NoDuplicateIDCheck.java
```diff
@@ -80,28 +85,83 @@ private void handleIdAttribute(TagNode node) {
...
+   * @param secondScopes the second conditional-host path
+   * @return false only when the paths diverge through known mutually exclusive hosts
+   */
+  private static boolean canCoexist(List<TagNode> firstScopes, List<TagNode> secondScopes) {
+    int sharedDepth = Math.min(firstScopes.size(), secondScopes.size());
+    for (int index = 0; index < sharedDepth; index++) {
+      TagNode firstScope = firstScopes.get(index);
+      TagNode secondScope = secondScopes.get(index);
+      if (firstScope != secondScope) {
+        return !TemplateConditionalScopeTracker.areMutuallyExclusive(firstScope, secondScope);
```
victor-peixoto-sonarsource: `canCoexist` returns at the first level where the two paths differ, so an exclusive divergence deeper in the path is never reached. Since `canCoexist == true` is the *raise* condition at the call site, a non-exclusive divergence at an outer level forces a report.

```html
<div *ngIf="isAdmin">
  <div *ngIf="editing"><input id="name"></div>
</div>
<div *ngIf="isUser">
  <div *ngIf="!editing"><input id="name"></div>
</div>
```

Paths are `[isAdmin, editing]` and `[isUser, !editing]`. At index 0 `isAdmin`/`isUser` are not mutually exclusive, so we return `true` and report — even though `editing`/`!editing` make the two `<input>`s impossible to render together. Ran this against the branch head:

```
Duplicate id "name" found. First occurrence was on line 5.  (line 8)
```

Suggest scanning the remaining shared indices for *any* exclusive pair instead of returning on the first mismatch:

```java
for (int index = 0; index < sharedDepth; index++) {
  TagNode firstScope = firstScopes.get(index);
  TagNode secondScope = secondScopes.get(index);
  if (firstScope != secondScope
    && TemplateConditionalScopeTracker.areMutuallyExclusive(firstScope, secondScope)) {
    return false;
  }
}
return true;
```

Worth noting this change is monotone in the safe direction: continuing the loop can only flip a `true` to a `false`, so it can only ever *remove* reports — it cannot introduce a new FP.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/TemplateConditionalScopeTracker.java
```diff
@@ -237,7 +251,133 @@ private int currentElementDepth() {
...
+   * @param secondScope the second conditional host
+   * @return whether the two hosts are mutually exclusive
+   */
+  public static boolean areMutuallyExclusive(TagNode firstScope, TagNode secondScope) {
+    return areOppositeAngularIfBranches(firstScope, secondScope)
+      || areVueConditionalBranches(firstScope, secondScope)
+      || areAngularSwitchBranches(firstScope, secondScope);
+  }
+
+  private static boolean areOppositeAngularIfBranches(TagNode firstScope, TagNode secondScope) {
```
victor-peixoto-sonarsource: This compares the raw `*ngIf` attribute value, but the NgIf microsyntax allows `as` aliases and `; else` / `; then` suffixes, so the most common two-branch Angular idiom doesn't match:

```html
<div *ngIf="cond; else other"><span id="x"></span></div>
<ng-template #other><span id="x"></span></ng-template>
```

`Condition.from("cond; else other")` normalises to the expression `cond;elseother`, which equals neither `cond` nor another suffixed form, so the pair is never recognised as opposite.

Two things compound here: the `<ng-template>` body also carries no conditional attribute at all, so its `id` goes through `registerUnconditionalId` and gets reported against the `*ngIf` branch.

Suggest trimming the microsyntax at the first `;` (and at ` as `) before comparing, and not treating an `<ng-template>` body referenced by `ngIfElse` / `ngSwitch` as unconditional.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/TemplateConditionalScopeTracker.java
```diff
@@ -33,18 +35,30 @@
  */
 public final class TemplateConditionalScopeTracker {
 
+  private static final String VUE_IF_ATTRIBUTE = "v-if";
+  private static final String VUE_ELSE_IF_ATTRIBUTE = "v-else-if";
+  private static final String VUE_ELSE_ATTRIBUTE = "v-else";
+  private static final String ANGULAR_IF_ATTRIBUTE = "*ngIf";
+  private static final String ANGULAR_SWITCH_CASE_ATTRIBUTE = "*ngSwitchCase";
+  private static final String ANGULAR_SWITCH_DEFAULT_ATTRIBUTE = "*ngSwitchDefault";
```
victor-peixoto-sonarsource, line 48: These three constants only match the `*`-prefixed sugar, while `isAngularSwitchHost` (line 369) accepts both `[ngSwitch]` and bare `ngSwitch`. That asymmetry means the de-sugared forms carry no recognised conditional branch attribute, so their descendants fall through to `registerUnconditionalId`:

```html
<div [ngSwitch]="x">
  <ng-template ngSwitchCase="'a'"><span id="p"></span></ng-template>
  <ng-template ngSwitchCase="'b'"><span id="p"></span></ng-template>
</div>
```

Same for `<ng-template ngSwitchDefault>` and `<ng-template [ngIf]="cond">`. Since the host form is already handled both ways, it'd be consistent to accept `ngIf`/`[ngIf]`, `ngSwitchCase`, and `ngSwitchDefault` here too.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/TemplateConditionalScopeTracker.java
```diff
@@ -237,7 +251,133 @@ private int currentElementDepth() {
...
+    String firstCase = firstScope.getAttribute(ANGULAR_SWITCH_CASE_ATTRIBUTE);
+    String secondCase = secondScope.getAttribute(ANGULAR_SWITCH_CASE_ATTRIBUTE);
+    return firstCase != null && secondCase != null && areDistinctLiteralValues(firstCase, secondCase);
+  }
+
+  private static boolean isAngularSwitchHost(TagNode node) {
+    return node.hasAttribute("[ngSwitch]") || node.hasAttribute("ngSwitch");
+  }
+
+  private static boolean areDistinctLiteralValues(String firstValue, String secondValue) {
```
victor-peixoto-sonarsource, line 545: Two issues with comparing raw attribute text through `QUOTED_LITERAL_PATTERN` (`['"].*['"]`, line 61):

1. **Over-claims exclusivity.** `*ngSwitchCase="'A'"` vs `*ngSwitchCase='"A"'` are "distinct literals" and declared exclusive, but `ngSwitch` renders *every* case matching the bound value — both branches render and the duplicate is missed. The pattern also accepts `'a' + suffix` and mismatched quotes (`"a'`) as literals.
2. **Under-claims exclusivity.** `*ngSwitchCase=" 1"` fails `NUMERIC_LITERAL_PATTERN.matches()` on the leading space, so no exclusivity is inferred and we report.

Trimming the value and tightening the quoted pattern to a real single-literal shape (matching quotes, no embedded unescaped delimiter, nothing trailing) would address both.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/TemplateConditionalScopeTracker.java
```diff
@@ -237,7 +251,133 @@ private int currentElementDepth() {
...
+    return firstBranchStart != null && firstBranchStart == vueConditionalBranchStart(secondScope);
+  }
+
+  @Nullable
+  private static TagNode vueConditionalBranchStart(TagNode scope) {
+    if (!hasVueConditionalAttribute(scope)) {
+      return null;
+    }
+    TagNode parent = scope.getParent();
+    if (parent == null) {
```
victor-peixoto-sonarsource, line 372: Bailing out when the host has no parent `TagNode` means a root-level `v-if`/`v-else` pair is never recognised as a chain. In a `.vue` file `<template>` supplies the parent (hence the passing fixture), but in a plain `.html` or an included fragment the top-level elements have `getParent() == null`:

```html
<div v-if="a"><span id="x"></span></div>
<div v-else><span id="x"></span></div>
```

Sibling detection could fall back to document order among top-level nodes rather than requiring a parent element.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/TemplateConditionalScopeTracker.java
```diff
@@ -237,7 +251,133 @@ private int currentElementDepth() {
   }
 
   public boolean isInConditional(TagNode node) {
```
victor-peixoto-sonarsource: Small. `isInConditional(TagNode)` now has no production caller (only the test uses it), and it allocates an `ArrayList` per call via `conditionalAttributeScopes`. Its semantics also changed silently — ancestors now count — for a class in the `...html.api` package. Worth either removing it or documenting the new meaning.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/coding/NoDuplicateIDCheck.java
```diff
@@ -43,11 +46,13 @@ public class NoDuplicateIDCheck extends AbstractPageCheck {
 
   // IDs seen outside any conditional - these are the "authoritative" IDs
   private final Map<String, Integer> unconditionalIds = new HashMap<>();
+  private final Map<String, List<ConditionalIdOccurrence>> conditionalIds = new HashMap<>();
```
victor-peixoto-sonarsource: `conditionalIds` retains `TagNode` references (and through them the whole parsed tree) until the next `startDocument`; `unconditionalIds` previously held only strings and ints. Fine for per-file analysis, just a heavier retention profile than before.

victor-peixoto-sonarsource (approved):
Approved with suggestions.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/coding/NoDuplicateIDCheck.java
erwan-leforestier-sonarsource: I've applied your suggestion

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/TemplateConditionalScopeTracker.java
erwan-leforestier-sonarsource: I've applied your suggestion

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/TemplateConditionalScopeTracker.java
erwan-leforestier-sonarsource: I've applied your suggestion

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/TemplateConditionalScopeTracker.java
erwan-leforestier-sonarsource: I've applied your suggestion

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/TemplateConditionalScopeTracker.java
erwan-leforestier-sonarsource: I've applied your suggestion

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/TemplateConditionalScopeTracker.java
erwan-leforestier-sonarsource: I've applied your suggestion

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/coding/NoDuplicateIDCheck.java
erwan-leforestier-sonarsource: I've switched stored scopes to lightweight values.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/TemplateConditionalScopeTracker.java
```diff
@@ -236,8 +281,264 @@ private int currentElementDepth() {
...
+      }
+      String conditionText = angularIfCondition(tagNode);
+      if (conditionText == null) {
+        continue;
+      }
+      Condition condition = Condition.fromAngularIf(conditionText);
+      registerAngularTemplateScope(angularIfTemplateReference(tagNode, conditionText, "then"), condition);
+      registerAngularTemplateScope(angularIfTemplateReference(tagNode, conditionText, "else"), condition.opposite());
+    }
+  }
```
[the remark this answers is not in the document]
erwan-leforestier-sonarsource: Fixed in 7b0bbb58. IDs in an ngIf host that explicitly delegates rendering to a then template are now ignored, since Angular renders the referenced template rather than that host body. Added a S7930 regression case; `mvn -pl sonar-html-plugin verify` passes (753 tests).

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/TemplateConditionalScopeTracker.java
```diff
@@ -236,8 +281,264 @@ private int currentElementDepth() {
...
+
```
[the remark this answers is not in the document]
erwan-leforestier-sonarsource: Fixed in 7b0bbb58. A template referenced by ngIf hosts with different conditions is now kept as an ambiguous conditional scope, so it is never assumed mutually exclusive with a branch. Added the S7930 regression case; `mvn -pl sonar-html-plugin verify` passes (753 tests).

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/TemplateConditionalScopeTracker.java
```diff
@@ -236,8 +281,264 @@ private int currentElementDepth() {
...
+      if (value != null) {
+        return value;
+      }
+    }
+    return null;
+  }
+
+  private static String scopeKey(TagNode node) {
+    return node.getStartLinePosition() + ":" + node.getStartColumnPosition();
+  }
```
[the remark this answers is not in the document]
erwan-leforestier-sonarsource: Not applicable to S7930: `NoDuplicateIDCheck` does not implement `EmbeddedHtmlCheck`, and `HtmlAstScanner` skips nodes with `isEmbedded()` for checks without that marker. Re-lexed PHP-string nodes therefore never reach this check, so coordinate collisions cannot affect this rule.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/TemplateConditionalScopeTracker.java
```diff
@@ -236,8 +280,304 @@ private int currentElementDepth() {
...
+
+  /**
+   * Returns whether the current element is inside an Angular {@code ngIf} host that delegates its
+   * rendered view to a {@code then} template.
+   *
+   * @return whether the element cannot be rendered because an explicit {@code then} template is used
+   */
+  public boolean isInsideAngularIfThenHost() {
+    return openElements.stream().anyMatch(TemplateConditionalScopeTracker::hasAngularThenTemplate);
+  }
```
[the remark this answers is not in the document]
erwan-leforestier-sonarsource: Fixed in 058a0829. The current node is now checked as well as open ancestors, so a void ngIf host with an explicit then template is suppressed even though it is not placed on the open-element stack. Added the void-host regression; `mvn -pl sonar-html-plugin verify` passes (753 tests).

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/TemplateConditionalScopeTracker.java
```diff
@@ -1482,6 +1824,114 @@ private boolean isInPersistentComment() {
...
+
+    private static AngularIfScope ambiguous() {
+      return new AngularIfScope(null, 0);
+    }
+
+    private boolean matches(Condition condition) {
+      return expression != null
+        && expression.equals(condition.expression())
+        && negationCount == condition.negationCount();
+    }
```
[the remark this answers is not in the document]
erwan-leforestier-sonarsource: Fixed in 058a0829. Template-condition equivalence now compares negation parity, matching the mutual-exclusion test, so `a` and `!!a` preserve a shared template scope instead of making it ambiguous. Added the regression; `mvn -pl sonar-html-plugin verify` passes (753 tests).

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/coding/NoDuplicateIDCheck.java
```diff
@@ -80,36 +86,113 @@ public void endElement(TagNode node) {
...
+    List<ConditionalAttributeScope> secondScopes) {
+    int sharedDepth = Math.min(firstScopes.size(), secondScopes.size());
+    for (int index = 0; index < sharedDepth; index++) {
+      ConditionalAttributeScope firstScope = firstScopes.get(index);
+      ConditionalAttributeScope secondScope = secondScopes.get(index);
+      if (TemplateConditionalScopeTracker.areMutuallyExclusive(firstScope, secondScope)) {
+        return false;
+      }
+    }
+    return true;
```
victor-peixoto-sonarsource, line 173: `canCoexist` compares the two host paths only at matching indices, but `conditionalAttributeScopes()` returns a *filtered* subsequence of the ancestor chain (only ancestors that are conditional hosts). So the paths line up only when the mutually exclusive hosts happen to sit at the same depth — otherwise the exclusive pair is never compared, and a duplicate is reported for markup that can never render twice.

Repro, run against this branch:

```html
<div *ngIf="loggedIn">
  <div *ngIf="isAdmin"><span id="misaligned-exclusive">Deep</span></div>
</div>
<div *ngIf="!isAdmin"><span id="misaligned-exclusive">Shallow</span></div>
```

The paths are `[loggedIn, isAdmin]` and `[!isAdmin]`, so `sharedDepth` is 1 and only `loggedIn` vs `!isAdmin` gets compared — different expressions, so "not exclusive":

```
Duplicate id "misaligned-exclusive" found. First occurrence was on line 12.
```

`isAdmin` / `!isAdmin` are mutually exclusive, so this is a false positive, and it reproduces in both nesting orders (deep-then-shallow and shallow-then-deep). Worth noting it is a regression against `master`: before this PR, descendants of conditional attributes were only compared against unconditional IDs and never against each other.

The existing fixture covers only the aligned case (`conditionalBlocksAngularDescendants.html:35-40`, where `editing` / `!editing` are both at depth 1), which is why this passes today.

Comparing every scope against every scope fixes it, and is safe: all three `areMutuallyExclusive` predicates require a matching identity key before returning true (same ngIf expression, same `vueBranchStartKey`, same `angularSwitchHostKey`), and for the Vue and switch cases a shared key implies a shared parent — so the cross product cannot invent an exclusivity that the index walk would have rejected.

```suggestion
    for (ConditionalAttributeScope firstScope : firstScopes) {
      for (ConditionalAttributeScope secondScope : secondScopes) {
        if (TemplateConditionalScopeTracker.areMutuallyExclusive(firstScope, secondScope)) {
          return false;
        }
      }
    }
    return true;
```

With this applied the repro reports nothing and `NoDuplicateIDCheckTest` + `TemplateConditionalScopeTrackerTest` stay green (90/90). Might be worth adding the misaligned case to `conditionalBlocksAngularDescendants.html` so it stays covered.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/coding/NoDuplicateIDCheck.java
```diff
@@ -80,36 +86,113 @@ public void endElement(TagNode node) {
   }
 
   private void handleIdAttribute(TagNode node) {
-    if (conditionalScope.isInNonRenderedRazorContent()) {
+    if (conditionalScope.isInNonRenderedRazorContent() || conditionalScope.isInsideAngularIfThenHost(node)) {
```
victor-peixoto-sonarsource, line 89: Minor (perf): `handleIdAttribute` is called from `startElement` for every element in every analyzed file, and this is the most expensive guard in the method — `isInsideAngularIfThenHost` streams the whole open-element stack, probing three attribute names and possibly running a regex per ancestor. It decides something that only matters for elements that actually carry an `id`, and the overwhelming majority do not, so that walk is computed and thrown away.

Both guards are side-effect-free reads, so hoisting the `id` lookup above them is behavior-preserving and skips the ancestor walk for nearly every element:

```java
String idValue = node.getAttribute("id");
if (shouldIgnoreId(idValue)) {
  return;
}
if (conditionalScope.isInNonRenderedRazorContent() || conditionalScope.isInsideAngularIfThenHost(node)) {
  return;
}
```

(Plain block rather than a `suggestion`: the reorder spans lines 89-95, but line 93 falls in the gap between two diff hunks, so a multi-line suggestion anchored there is not accepted. Verified green with the reorder applied.)

victor-peixoto-sonarsource (approved):
Approved with suggestions.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/coding/NoDuplicateIDCheck.java
erwan-leforestier-sonarsource: Good catch — the conditional scope lists are filtered ancestor paths, so equal list indexes do not reliably represent the same nesting depth. I now compare every pair of scopes and added the misaligned Angular `isAdmin` / `!isAdmin` regression in both source orders. `NoDuplicateIDCheckTest` and `TemplateConditionalScopeTrackerTest` pass (90 tests) in 2b42963f.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/coding/NoDuplicateIDCheck.java
erwan-leforestier-sonarsource: Applied. The `id` lookup and cheap ignore check now run before the Razor/Angular ancestor checks, avoiding the `isInsideAngularIfThenHost` walk for elements without a usable `id`. Verified with `NoDuplicateIDCheckTest` and `TemplateConditionalScopeTrackerTest` (90 tests) in 2b42963f.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/TemplateConditionalScopeTracker.java
```diff
@@ -236,8 +280,305 @@ private int currentElementDepth() {
...
+      || areAngularSwitchBranches(firstScope, secondScope);
+  }
+
+  private static boolean areOppositeAngularIfBranches(
+    ConditionalAttributeScope firstScope,
+    ConditionalAttributeScope secondScope) {
+    return firstScope.angularIfExpression() != null
+      && firstScope.angularIfExpression().equals(secondScope.angularIfExpression())
+      && firstScope.angularIfNegationCount() % 2 != secondScope.angularIfNegationCount() % 2;
+  }
```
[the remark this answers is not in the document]
erwan-leforestier-sonarsource: Fixed in 2b588594. `Condition.from` now preserves a leading `!` as a condition-level negation only when no top-level `&&` or `||` follows it. This keeps `!(a || b)` and `a || b` mutually exclusive, while reporting the real duplicate for `!a || b` and `a || b`. Added both regression cases in `conditionalBlocksAngularParenthesizedConditions.html`; `mvn -pl sonar-html-plugin verify` passes (753 tests).

PR #832, opened 2026-09-11
SONARHTML-492 Improve rule S1101

guillemsarda:
Part of 

----
## Summary by Gitar

- **Rule improvements:**
    - Enhanced `LinksIdenticalTextsDifferentTargetsCheck` to support accessibility attributes like `aria-label` and `aria-labelledby`, and ignore hidden links
    - Updated rule documentation and configuration for S1101


sonar-html-plugin/src/main/resources/org/sonar/l10n/web/rules/Web/LinksIdenticalTextsDifferentTargetsCheck.json
```diff
@@ -1,10 +1,17 @@
 {
   "title": "Links with identical texts should have identical targets",
   "type": "CODE_SMELL",
+  "code": {
+    "impacts": {
+      "MAINTAINABILITY": "LOW",
+      "RELIABILITY": "LOW"
+    },
+    "attribute": "CONVENTIONAL"
+  },
```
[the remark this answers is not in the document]
guillemsarda: Good catch, applying the fix in the corresponding rspec PR: https://github.com/SonarSource/rspec/pull/8185.

sonar-html-plugin/src/main/resources/org/sonar/l10n/web/rules/Web/LinksIdenticalTextsDifferentTargetsCheck.html
```diff
@@ -5,30 +5,20 @@ <h2>Why is this an issue?</h2>
 each link. Links in separate containers (different <code>&lt;article&gt;</code>, <code>&lt;li&gt;</code>, <code>&lt;div&gt;</code>, or other elements)
 are intentionally not flagged: WCAG 2.4.4 explicitly allows identical link text when the surrounding context (for example, the enclosing list item or
 table row) provides disambiguation.</p>
-<h3>Noncompliant code example</h3>
-<pre>
+<h3>Code examples</h3>
+<h4>Noncompliant code example</h4>
+<pre data-diff-id="1" data-diff-type="noncompliant">
```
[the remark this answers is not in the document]
guillemsarda: It's fine. Also, this is handled in another PR and the content will be mirrored from there.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/LinksIdenticalTextsDifferentTargetsCheck.java
```diff
@@ -41,54 +50,80 @@ public class LinksIdenticalTextsDifferentTargetsCheck extends AbstractPageCheck
 
   @Override
   public void startDocument(List<Node> nodes) {
-    linksByParent.clear();
+    unconditionalLinksByParent.clear();
     inLink = false;
+    conditionalScope.reset(Helpers.isRazorFile(getHtmlSourceCode()));
```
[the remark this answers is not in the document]
guillemsarda: Out of scope for the rspec PR (SonarSource/rspec#8185) — this is a Java-level wiring bug in `TemplateConditionalScopeTracker`/the check itself, not a rule-description issue. Needs a code fix directly in this PR.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/LinksIdenticalTextsDifferentTargetsCheck.java
```diff
@@ -41,54 +50,80 @@ public class LinksIdenticalTextsDifferentTargetsCheck extends AbstractPageCheck
...
+
+    Map<String, Link> siblingLinks = unconditionalLinksByParent.computeIfAbsent(linkParent, k -> new HashMap<>());
+    Link previousLink = siblingLinks.get(upperText);
+    if (previousLink == null) {
+      registerBaseline(siblingLinks, upperText);
+    } else if (!target.equals(previousLink.getTarget())) {
+      createViolation(line, "Use distinct texts or point to the same target for this link and the one at line " + previousLink.getLine() + ".");
+      registerBaseline(siblingLinks, upperText);
+    }
+  }
```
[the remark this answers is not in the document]
guillemsarda: Out of scope for the rspec PR (SonarSource/rspec#8185) — the doc-facing side of this (documenting that only earlier unconditional links form the baseline) is covered there via the fix for the sibling comment below. The underlying order-dependent behavior itself is a code matter for this PR, not something to change from the rule description.

sonar-html-plugin/src/test/resources/checks/LinksIdenticalTextsDifferentTargetsCheck.cshtml
```diff
@@ -0,0 +1,52 @@
...
+  <a href="/offer/2">Offer</a>
+</div>
+
+<!-- Compliant: link inside a C# line comment inside a Razor code block is not rendered, so it is ignored -->
+@{
+    // <a href="/deal/1">Deal</a>
+}
+<div>
+  <a href="/deal/2">Deal</a>
+</div>
```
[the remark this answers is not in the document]
guillemsarda: Out of scope for the rspec PR (SonarSource/rspec#8185) — this is a test-fixture correctness issue in `LinksIdenticalTextsDifferentTargetsCheck.cshtml`, unrelated to the rule description content. Needs a fix directly in this PR.

sonar-html-plugin/src/main/resources/org/sonar/l10n/web/rules/Web/LinksIdenticalTextsDifferentTargetsCheck.html
```diff
@@ -5,30 +5,30 @@ <h2>Why is this an issue?</h2>
 each link. Links in separate containers (different <code>&lt;article&gt;</code>, <code>&lt;li&gt;</code>, <code>&lt;div&gt;</code>, or other elements)
 are intentionally not flagged: WCAG 2.4.4 explicitly allows identical link text when the surrounding context (for example, the enclosing list item or
 table row) provides disambiguation.</p>
-<h3>Noncompliant code example</h3>
-<pre>
+<h3>Exceptions</h3>
+<p>This rule only considers <code>&lt;a&gt;</code> elements.</p>
+<p>A link with no accessible name at all is not considered by this rule; see {rule:Web:S6827}.</p>
```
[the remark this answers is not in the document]
guillemsarda: This and comment https://github.com/SonarSource/sonar-html/pull/832#discussion_r4014310021 fall into the RSPEC PR. The content of this file will be mirrored automatically. Do not review again

sonar-html-plugin/src/main/resources/org/sonar/l10n/web/rules/Web/LinksIdenticalTextsDifferentTargetsCheck.html
```diff
@@ -5,30 +5,30 @@ <h2>Why is this an issue?</h2>
...
+<p>A link that is hidden from every user, for example with <code>aria-hidden="true"</code>, the <code>hidden</code> attribute, or
+<code>style="display: none"</code>, is not considered by this rule either, since it cannot cause confusion for anyone.</p>
+<p>In Razor files (<code>.cshtml</code>, <code>.vbhtml</code>), links inside a Razor comment or a C# comment are not considered either, since they are
+never rendered.</p>
+<p>A link inside a conditional block (for example, an <code>@if</code>/<code>@else</code> pair, a <code>v-if</code>/<code>v-else</code> pair, or a
+JSTL <code>&lt;c:if&gt;</code>) is only compared against links that appear outside any conditional block in the same parent. Two links that both sit
+inside conditional blocks are therefore never flagged against each other, even when their branches are independent rather than mutually exclusive.</p>
```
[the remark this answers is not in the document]
guillemsarda: Not fixing it, is RSPEC scope

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/LinksIdenticalTextsDifferentTargetsCheck.java
```diff
@@ -41,54 +50,80 @@ public class LinksIdenticalTextsDifferentTargetsCheck extends AbstractPageCheck
...
-        } else {
-          siblingLinks.put(upperText, new Link(line, target));
-        }
+      if (!linkHidden) {
+        recordOrCompareLink();
       }
     }
   }
 
+  private void recordOrCompareLink() {
```
martin-strecker-sonarsource, line 124: **[blocker]** A conditional link that appears *before* its unconditional sibling in document order is never compared against it, because `registerBaseline()` drops conditional links outright rather than retaining them as a pending candidate the way `NoDuplicateIDCheck` (the stated reference pattern) does.

Example: `<div>@if(x){<a href="/a">Edit</a>}<a href="/b">Edit</a></div>` — the conditional link finds no previous entry and is dropped; the unconditional link then also finds no previous entry and simply becomes the baseline. No violation is raised, even though both links render together whenever `x` is true and point to different targets. This is the same scenario the test fixture covers in the opposite order.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/LinksIdenticalTextsDifferentTargetsCheck.java
```diff
@@ -41,54 +50,80 @@ public class LinksIdenticalTextsDifferentTargetsCheck extends AbstractPageCheck
...
+    if (previousLink == null) {
+      registerBaseline(siblingLinks, upperText);
+    } else if (!target.equals(previousLink.getTarget())) {
+      createViolation(line, "Use distinct texts or point to the same target for this link and the one at line " + previousLink.getLine() + ".");
+      registerBaseline(siblingLinks, upperText);
+    }
+  }
+
+  private void registerBaseline(Map<String, Link> siblingLinks, String upperText) {
+    if (!linkInConditional) {
```
martin-strecker-sonarsource: **[blocker]** Two links inside the *same* conditional branch — guaranteed to render together — are never compared against each other either, since `registerBaseline()` only stores a link when `!linkInConditional`, so neither ever becomes a comparison anchor.

Example: `<div>@if(x){<a href="/a">Edit</a><a href="/b">Edit</a>}</div>` — a guaranteed-coexisting pair with identical text and different targets is silently missed.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/AccessibilityUtils.java
```diff
@@ -118,6 +121,21 @@ public static boolean isHiddenFromScreenReader(TagNode element) {
...
   }
 
+  /**
+   * Returns whether {@code element} is unconditionally hidden from every user via the native
+   * {@code hidden} boolean attribute or an inline {@code style="display: none"}. Unlike
+   * {@link #isHiddenFromScreenReader}, this does not inspect CSS classes or computed styles —
+   * only the two purely syntactic, unambiguous signals.
+   */
+  public static boolean isHiddenByDisplayNone(TagNode element) {
+    return element.hasAttribute("hidden") || DISPLAY_NONE_PATTERN.matcher(getStyleOrEmpty(element)).find();
```
martin-strecker-sonarsource: **[suggestion]** `isHiddenByDisplayNone` uses plain `hasAttribute`/`getAttribute` for `hidden`/`style`, instead of the property-binding-aware `getPropertyValue` that `isHiddenFromScreenReader` uses two methods above it in this same file for `aria-hidden`/`type`. An Angular/Vue-bound `[hidden]="true"` or `[attr.style]="'display:none'"` link isn't recognized as hidden, so it gets compared normally and can produce a spurious violation — the exact FP class this ticket set out to eliminate.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/AccessibilityUtils.java
```diff
@@ -64,6 +65,8 @@ public class AccessibilityUtils {
     .map(name -> name.toLowerCase(Locale.ROOT))
     .collect(Collectors.toUnmodifiableSet());
 
+  private static final Pattern DISPLAY_NONE_PATTERN = Pattern.compile("display\\s*:\\s*none", Pattern.CASE_INSENSITIVE);
```
martin-strecker-sonarsource: **[nitpick]** `DISPLAY_NONE_PATTERN` matches the *first* `display:none` occurrence even if a later declaration in the same `style` attribute overrides it. `style="display:none; display:block"` is actually visible per CSS cascade, but gets wrongly excluded from comparison. Narrow edge case, low real-world likelihood.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/LinksIdenticalTextsDifferentTargetsCheck.java
martin-strecker-sonarsource, line 66: **[question]** This calls `conditionalScope.reset(razorFile)`, but the stated reference pattern (`NoDuplicateIDCheck`) uses the `reset(razorFile, nodes)` overload. As a result, Angular `ngIf` then/else `ng-template` scopes are never registered here. It's currently masked because `ng-template` siblings already have distinct parent identities, but it's a real divergence from the reference pattern — was this intentional, or should it match `NoDuplicateIDCheck`'s call?

martin-strecker-sonarsource (changes requested):
Clean rollback of S9384 (verified: no leftover references in code, tests, metadata, or the profile), and unit tests pass. But the conditional-branch baseline logic has two ordering bugs that undercut the ticket's core deliverable, plus a hidden-link detection gap and a divergence from the stated reference pattern worth a second look.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/LinksIdenticalTextsDifferentTargetsCheck.java
guillemsarda: Good catch, but confirmed harmless here — every comparison in this rule is scoped by `linkParent` (the link's direct DOM parent). A `<ng-template>` is always a distinct parent from anything outside it, so a link inside `tplA`/`tplB` can never share a parent with a link outside those templates — it's simply never compared either way, whether or not the tracker recognizes the template as conditional. And two links inside the *same* `<ng-template>` genuinely do render together, so treating them as "unconditional" relative to each other is still correct without the extra registration.

That's different from `NoDuplicateIDCheck`, which compares IDs document-wide (not parent-scoped) — that's exactly why *it* needs the Angular template registration, to avoid flagging `then`/`else` duplicates as real ones.

Given that, keeping `reset(razorFile)` here rather than switching to the `nodes` overload — it wouldn't change any observable behavior in this rule, just add unused plumbing.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/LinksIdenticalTextsDifferentTargetsCheck.java
guillemsarda: Good catch, fixed in 8c1041d — conditional links are now kept as pending candidates and compared once an unconditional sibling appears, regardless of order.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/LinksIdenticalTextsDifferentTargetsCheck.java
guillemsarda: Fixed in e9d16174 (extra coverage in 7119693d) — links now compare against a same-branch sibling via branch identity, without conflating mutually exclusive branches.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/AccessibilityUtils.java
guillemsarda: Fixed in 9a5bf00e — hidden/style detection now goes through the property-binding-aware lookup, matching aria-hidden's pattern.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/AccessibilityUtils.java
guillemsarda: Fixed in 9b5fbc09 — now takes the last display declaration per the CSS cascade instead of the first match.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/accessibility/AccessibilityUtils.java
```diff
@@ -112,12 +117,76 @@ public static boolean isHiddenFromScreenReader(TagNode element) {
...
+    return lastDisplayValue != null && DISPLAY_NONE_VALUE_PATTERN.matcher(lastDisplayValue).find();
+  }
+
+  /**
+   * Returns the value of the last {@code property: value} declaration in a semicolon-separated CSS
+   * declaration list, honoring the cascade rule that a later declaration overrides an earlier one
+   * for the same property. Returns {@code null} when {@code property} is absent.
+   */
+  @Nullable
+  private static String lastDeclarationValue(String style, String property) {
```
martin-strecker-sonarsource: **[suggestion]** I don't like this hand-parsing of CSS. `lastDeclarationValue`/`getStyleOrEmpty` is a bespoke, partial re-implementation of the CSS Syntax Module Level 3 tokenizer + "parse a list of declarations" algorithm (https://www.w3.org/TR/css-syntax-3/#parse-a-list-of-declarations -- this is the actual spec entry point HTML uses for style="" attribute content), and it's missing real gotchas that algorithm defines:

- **Comments** (slash-star ... star-slash) aren't stripped before tokenizing, so a semicolon or colon inside a CSS comment would confuse the current split(";") approach.
- **String escapes and bad-string recovery**: a backslash-escaped quote inside a quoted value, or a raw newline inside a string (which the spec says should produce a bad-string-token and recover, not silently corrupt the rest of the declaration list), aren't handled.
- **url(...) tokens** have their own consumption path, separate from quoted strings, with different escaping/whitespace rules and their own bad-url recovery routine -- a url() value isn't parsed the way a quoted string is.
- **Escaped code points**: a backslash followed by hex digits (or a literal char) is a general CSS escape mechanism, independent of quoting, that nothing here accounts for.
- **Numeric token grammar** (numbers vs. percentages vs. dimensions, scientific notation, signs) -- not relevant to *this* check's `display` lookup today, but it's part of what real CSS parsing means, and it's the kind of thing that quietly breaks if this helper ever gets reused for another property.

Two ways to actually solve this instead of patching one more edge case at a time:

1. **Use `com.helger:ph-css`** (Apache-2.0, actively maintained -- v8.2.1 as of 2026-05-16, used inside Apache JMeter). `CSSReaderDeclarationList.readFromString(styleValue, ECSSVersion.CSS30)` implements exactly the "parse a list of declarations" entry point above, correctly, including all of the gotchas listed. This would be the first CSS-parsing dependency in this codebase, but it's a well-fitted, purpose-built one for precisely this problem.
2. **Write a proper `CssStyleLexer.java`**, held to the same bar as `PageLexer.java` handles HTML: a real tokenizer covering the gotchas above (comments, string/bad-string recovery, url()/bad-url recovery, escaped code points), not a split(";") helper with a docstring. This is a genuinely non-trivial amount of code to get right and keep right -- worth it only if we expect to reuse CSS parsing beyond this one `display` lookup.

Either is better than continuing to hand-patch `lastDeclarationValue` one discovered edge case at a time -- right now we're accumulating exactly the kind of narrow, self-written parser risk this comment is trying to avoid.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/checks/sonar/LinksIdenticalTextsDifferentTargetsCheck.java
```diff
@@ -41,52 +60,153 @@ public class LinksIdenticalTextsDifferentTargetsCheck extends AbstractPageCheck
...
   @Override
   public void startElement(TagNode node) {
+    conditionalScope.startElement(node);
     if (isA(node)) {
       inLink = true;
       text.delete(0, text.length());
       target = getTarget(node);
       line = node.getStartLinePosition();
       linkParent = node.getParent();
+      linkHidden = isHiddenLink(node) || conditionalScope.isInNonRenderedRazorContent();
```
martin-strecker-sonarsource: **[blocker]** `isHiddenLink`/`isHiddenFromScreenReader`/`isHiddenByDisplayNone` only inspect the anchor node itself -- I grepped `AccessibilityUtils.java` for `getParent` and got zero matches, and re-read `startElement`/`endElement` here byte-for-byte: neither branches on anything except `isA(node)`, so a wrapping element passing through this visitor never updates any hidden-tracking state. `TemplateConditionalScopeTracker` (the other piece of shared per-element state) has zero references to hidden/display/aria-hidden either -- it is exclusively about template conditional branches, a separate concern.

Concrete FP this causes: `<div style="display:none"><a href="/a">Edit</a></div><div><a href="/b">Edit</a></div>` -- the first `<a>` carries no hidden-related attribute itself, so it is compared normally against the second and gets flagged, even though it is invisible to every user because its *parent* is `display:none`. Per WAI-ARIA 1.2 section 7.1 (Excluding Elements from the Accessibility Tree, discussed earlier on rspec#8185), hidden-by-CSS propagates to every descendant, not just the element carrying the attribute -- this is exactly the FP class this ticket set out to eliminate, just one level up the tree.

The JS/TSX facet (SonarJS PR #7949) already has the fix, from an earlier round of this same review: `isHiddenByAncestor`, which walks `anchor.parent` and checks every ancestor `JSXElement` with the same hidden predicate used on the anchor itself. Suggest porting the same shape here: walk `node.getParent()` up from the anchor (or from `linkParent`) checking `isHiddenLink`-equivalent logic on each `TagNode` ancestor, stopping at the document root.

sonar-html-plugin/src/main/java/org/sonar/plugins/html/api/TemplateConditionalScopeTracker.java
```diff
@@ -1830,10 +1853,13 @@ private static final class ConditionalAttributeState {
     private final Map<String, AngularIfScope> angularTemplateScopes = new HashMap<>();
     @Nullable
     private String rootVueConditionalBranchStartKey;
+    // Shared bookkeeping, unrelated to conditional attributes: identifies the current text/JSTL branch.
+    private long conditionalBranchGeneration;
```
martin-strecker-sonarsource: **[suggestion]** I do not like this extra state variable -- `conditionalBranchGeneration` looks placed at the wrong spot.

It lives inside `ConditionalAttributeState`, which reads as "state about attribute-based conditionals (*ngIf, v-if, ...)". But it is mutated from six call sites, and only one of them (the JSTL tag check in the outer class's `startElement`) is even tag/attribute-adjacent -- the other five are the brace-based text-scanner's open/continue paths (Razor/Twig/PHP/Angular-brace), which have nothing to do with attributes. Meanwhile the class's actual attribute-conditional machinery -- `areMutuallyExclusive`/`areOppositeAngularIfBranches`/`areVueConditionalBranches`/`areAngularSwitchBranches` -- never touches this field at all. So the field is reached into from code that conceptually does not belong to its owning class, while the owning class's own comparison logic ignores it entirely.

This is not just cosmetic -- it hides a real gap. `LinksIdenticalTextsDifferentTargetsCheck` never calls `areMutuallyExclusive`; it only checks `attributeScopes.isEmpty()` and, whenever it is not empty, forces `linkBranchId = null`. In `recordOrCompareLink`, the read path is `linkBranchId == null ? null : pendingBranches.get(linkBranchId)` -- it short-circuits before ever attempting a lookup. So two links that share the exact same conditional-attribute host -- which is guaranteed to render together -- are never compared against each other. Concrete repro: `<div *ngIf="x"><a href="/a">Edit</a><a href="/b">Edit</a></div>` -- same parent, same *ngIf host, same accessible name, different targets, never flagged. This is the same bug class as the same-branch fix already applied for brace-based conditionals earlier in this review, just unfixed for the attribute-based family.

Proposed architecture, in two parts:

**1. Give the tracker one uniform branch-identity model instead of four bespoke comparison methods.** A single value type per open branch, used regardless of which engine opened it:

```java
record BranchScope(
  Object identity,                 // fresh per branch-opening event; replaces conditionalBranchGeneration,
                                    // minted locally at each open/continue site -- no shared field needed
  @Nullable Object constructKey,   // populated whenever the source structure alone determines "same construct":
                                    //   sequential generation for brace/tag-based chains,
                                    //   sibling-adjacency key for Vue v-if/v-else-if/v-else,
                                    //   switch-host key for Angular [ngSwitch]/*ngSwitchCase
  @Nullable ConditionSignature condition  // populated only when no structural anchor exists at all:
                                           //   normalized expression + negation parity, for a bare, unlinked *ngIf pair
)
```

with one generic decision procedure instead of `areOppositeAngularIfBranches`/`areVueConditionalBranches`/`areAngularSwitchBranches`: same `constructKey`, different `identity` -> mutually exclusive (covers brace-based chains, Vue, and Angular switch). No shared `constructKey`, but both scopes carry a `condition` and they are structurally complementary -> mutually exclusive (this is where the *ngIf="x"/*ngIf="!x" special case stays -- it genuinely has no structural anchor to lean on, so it is the one case that legitimately needs the text/negation comparison, kept as a narrow, explicit fallback rather than spread across bespoke per-engine methods). Otherwise -> unknown, do not compare.

**2. Stop tracking baseline/pending state incrementally in the check -- collect during the traversal, compare once in `endDocument()`.** `unconditionalLinksByParent` and `pendingConditionalLinksByParent` exist because the check currently has to decide, incrementally and in document order, whether each link is a reliable "baseline" or an uncertain "pending" candidate -- and remember pending ones in case a later link retroactively confirms coexistence. That complexity is only needed because the check reports while it visits. There is already a precedent in this codebase for not doing that: `IndistinguishableSimilarElementsCheck` buffers a plain `List<TagNode>` during `startElement` and does all its comparison in `endDocument()` (an existing, real override point on `DefaultNodeVisitor`). The same restructuring works here:
- `startElement`/`endElement`/`characters`/`directive` stay exactly as they are today (driving the tracker in document order is inherently sequential -- it is scanning raw text character-by-character), but instead of comparing anything inline, each anchor is just appended to a flat `List<LinkRecord(parent, nameKey, target, branchScope, line)>`.
- `endDocument()` groups that list by `(parent, nameKey)`, partitions each group into unconditional vs. conditional anchors (already known per record, no lookahead needed), and compares: every anchor with a conflicting target against any unconditional anchor in its group is a violation; conditional anchors are additionally compared pairwise only within groups whose `BranchScope`s are coexistence-compatible.

With everything collected upfront there is no more "baseline gets replaced on conflict, then re-registered" state machine and no more nested `Map<Object, Link>` lookup keyed by branch id -- just filter, group, and compare over an already-complete list. `unconditionalLinksByParent` and `pendingConditionalLinksByParent` disappear from the check entirely; `conditionalScope` (the tracker reference) stays, since something still has to produce each anchor's `BranchScope` during the collection pass.

That said, this touches the tracker's core model and `NoDuplicateIDCheck` is the other consumer, so it is bigger than this ticket's stated scope -- worth its own follow-up ticket rather than blocking this PR, with the concrete repro added in this PR marked as a bug with a link to that ticket (could you open one and link it here?).

martin-strecker-sonarsource (changes requested):
Following up on the comments just posted -- the ancestor-hidden gap is a real functional blocker (concrete FP repro in that comment).

PR #846, opened 2026-09-17
CLP-940 Make matrix fail-fast and required-check policy explicit

frederic-tingaud-sonarsource:
Part of CLP-907

Adds `fail-fast: false` to the `analyze-shadows` matrix in `build.yml` (2 cells: `sqc-eu`, `sqc-us`).

## Why

Making the policy explicit rather than inherited. If one platform's shadow analysis fails — usually a platform-side outage rather than a code problem — the other is currently cancelled.

**Being straight about the evidence: this is a consistency change, not a fix for a problem we have.** The job is gated `if: github.event_name == 'schedule'`, so it runs nightly only, never on PRs or master pushes. Last 100 `build.yml` runs (81 success / 16 failure / 2 cancelled / 1 action_required) contain **0 cancellations in this matrix**, and because it's nightly a lost cell is recovered by the next run ~24h later.

It is included for a uniform policy across the CLP-907 repos. If the ticket is being trimmed, this is a fair candidate to drop — please say so and I'll close it.

## Cost

Nothing on PR or master runs, since the job is schedule-gated. One extra analysis job on a failing nightly.

## Not included

No required-check changes — no matrix cell is a required status check in this repo (only the aggregate `StepSecurity Required Checks`, ruleset `21087917`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mary-georgiou (approved):
LGTM

